PR #4, opened 2023-01-23
docs: fix typo

qu3vipon:
fix a typo in the type hint

### DIFF
--- a/pcweb/pages/docs/advanced_guide/wrapping_react.py
+++ b/pcweb/pages/docs/advanced_guide/wrapping_react.py
@@ -43,7 +43,7 @@ class ColorPicker(pc.Component):
     color: pc.Var[str]
 
     @classmethod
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/advanced_guide/wrapping_react.py
```diff
@@ -43,7 +43,7 @@ class ColorPicker(pc.Component):
     color: pc.Var[str]
 
     @classmethod
-    def get_controlled_triggers(cls) -> set[str]:
+    def get_controlled_triggers(cls) -> Set[str]:
```
picklelo, line 46: This is valid from Python 3.9+ so we don't have to change it

qu3vipon: Fair enough. I thought it was a typo because pynecone repo uses old syntax.

PR #22, opened 2023-02-23
Add no-zip Parameter Flag to Documenation

r0b2g1t:
This PR add the no-zip feature to the documentation with an description.

### DIFF
--- a/pcweb/pages/docs/hosting/self_hosting.py
+++ b/pcweb/pages/docs/hosting/self_hosting.py
@@ -98,8 +98,8 @@ def self_hosting():
             pc.code("--no-zip"),
             " parameter.",
             "This provides the frontend in the ",
-            pc.code("_static/"),
-            " directory and the backend can be found in the root directory of the project.",
+            pc.code(".web/_static/"),
+            " directory and the backend can be found in the root directory of the project. ",
         ),
         subheader("Pynecone Container Service"),
         doctext(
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -93,6 +93,14 @@ def self_hosting():
             pc.code("--backend-only"),
             " flags.",
         ),
+        doctext(
+            "It is also possible to export the components without zip it separate. To do this, use the ",
+            pc.code("--no-zip"),
+            " parameter.",
```
picklelo, line 99: Can we add a space after the `.` here. We need a way to catch this automatically as it's easy to miss.

pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -93,6 +93,14 @@ def self_hosting():
...
+            "This provides the frontend in the ",
+            pc.code("_static/"),
```
picklelo, line 101: Should this be the `.web/_static/` directory?

picklelo (commented):
Thanks for adding docs! Super helpful - just small comments

pcweb/pages/docs/hosting/self_hosting.py
r0b2g1t: I have add the space

pcweb/pages/docs/hosting/self_hosting.py
r0b2g1t: I have add .web/ in front of _static/

picklelo (approved):
nice!

PR #23, opened 2023-02-26
Update self_hosting.py

funprogrammer89:
Added a space after the period in this line, 'Another option is to run your Pynecone service in a container.For this purpose, a'...

### DIFF
--- a/pcweb/pages/docs/hosting/self_hosting.py
+++ b/pcweb/pages/docs/hosting/self_hosting.py
@@ -104,7 +104,7 @@ def self_hosting():
         subheader("Pynecone Container Service"),
         doctext(
             "Another option is to run your Pynecone service in a container.",
-            "For this purpose, a ",
+            " For this purpose, a ",
             pc.code("Dockerfile"),
             " and additional documentation is available in the Pynecone project in the directory ",
             pc.code("docker-example"),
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -104,7 +104,7 @@ def self_hosting():
         subheader("Pynecone Container Service"),
         doctext(
             "Another option is to run your Pynecone service in a container.",
```
r0b2g1t, line 106: It is better to add the space in this line after the ```... container. "```
This is the way how the parser works.

r0b2g1t (changes requested):

pcweb/pages/docs/hosting/self_hosting.py
funprogrammer89: Should I create a new pull request with the correction handled the way you suggested? Thanks!

r0b2g1t: No, you can fix it, commit and push it in the same branch. That will add it automatically to the PR here. :)

funprogrammer89: Awesome, will do, thanks!

PR #24, opened 2023-03-02
Needed space after period in sentence.

funprogrammer89:
Added a space after the period in this line, ... Another option is to run your Pynecone service in a container. 

So that there is a space between the two sentences. =)

Altered my original commit per r0b2g1t's request.


r0b2g1t (approved):
Looks good

picklelo (approved):
nice catch!

PR #54, opened 2023-04-16
forms: update example for multi-file upload

masenf:
Multi-file upload support was added in pynecone-io/pynecone#712.

* Remove sentence saying that only single file upload is supported
* Add sentence explaining the an event handler argument with `list[UploadFile]` annotation will be used to process multi-upload
* Update (and test) upload example 2 to allow uploading multiple images, and displaying them all in a grid.
  * Include "accepts" prop in the example, since it's not clear from the docs what format is expected. (This can be updated over the in main pynecode repo eventually)

### DIFF
--- a/pcweb/pages/docs/component_lib/forms.py
+++ b/pcweb/pages/docs/component_lib/forms.py
@@ -840,16 +840,14 @@ def render_upload():
         ),
         doctext(
             "Your event handler should be an async function that accepts a single argument, ",
-            pc.code("file"),
-            ", which is a ",
+            pc.code("files: list[UploadFile]"),
+            ", which will contain ",
             doclink(
                 "FastAPI UploadFile",
                 "https://fastapi.tiangolo.com/tutorial/request-files",
             ),
-            ". Alternatively, the handler may accept a ",
-            pc.code("list[UploadFile]"),
-            "argument to process a multi-file upload. ",
-            "You can read the file and save it anywhere as shown in the example. ",
+            " instances. ",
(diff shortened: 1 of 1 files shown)


masenf:
CC: @Alek99 @picklelo @ElijahAhianyo who were involved in the feature.

pcweb/pages/docs/component_lib/forms.py
```diff
@@ -821,9 +838,6 @@ def render_upload():
             width="100%",
             padding_bottom="1em",
         ),
-        doctext(
-            "Currently we only support one file upload, but we plan to support multiple file uploads in the future. ",
-        ),
         doctext(
             "Your event handler should be an async function that accepts a single argument, ",
             pc.code("file"),
```
ElijahAhianyo, line 843: we've changed the syntax of the file uploads to use lists only for uniformity reasons. This means single uploads would use multi-upload syntax(`list[UploadFile]`). In the case of single uploads, the `multiple prop` should be set to `False`.

more [here](https://github.com/pynecone-io/pynecone/pull/811)

ElijahAhianyo (changes requested):
Nice. Thanks for the addition. Just some minor comments about syntax

pcweb/pages/docs/component_lib/forms.py
masenf: Thanks. Fixed.

ElijahAhianyo (approved):
lgtm

Alek99 (approved):
Nice thanks!

PR #59, opened 2023-04-24
Datatable docs modification:

ElijahAhianyo:
This PR adds the following changes:
- Fixes assertion error when ran on windows due to the inspect module using `\n` for linesep whereas its `\r\n` for windows.
- Picks up prop comments that span multiple line. Previously, only the line above the prop was used.
- Added more examples to datatable component docs.

### DIFF
--- a/pcweb/pages/docs/component.py
+++ b/pcweb/pages/docs/component.py
@@ -42,7 +42,7 @@ class Source(Base):
         # Get the source code.
         self.code = [
             line
-            for line in inspect.getsource(self.component).split("\n")
+            for line in inspect.getsource(self.component).splitlines()
             if len(line) > 0
         ]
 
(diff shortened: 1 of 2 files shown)


masenf (approved):
awesome 😁

pcweb/pages/docs/component.py
```diff
@@ -44,7 +42,7 @@ def __init__(self, *args, **kwargs):
         # Get the source code.
         self.code = [
             line
-            for line in inspect.getsource(self.component).split(os.linesep)
+            for line in inspect.getsource(self.component).split("\n")
```
masenf, line 45:
```suggestion
            for line in inspect.getsource(self.component).splitlines()

splitlines works on any type of line ending regardless of os (and might be more future-proof if python _does_ start returning os-specific line endings)

Alek99 (approved):
Lg

PR #72, opened 2023-05-16
Update LoggingMiddleware example

TudorAndrei-Pythia:
Update the postprocess method arguments to avoid the following error:

```
TypeError: LoggingMiddleware.postprocess() got an unexpected keyword argument 'update'
```


picklelo:
Thanks for catching! Merged #76 which did the same change

PR #83, opened 2023-05-25
Fix issue 82

milochen0418:
Fix the issue 82
- https://github.com/pynecone-io/pcweb/issues/82

### DIFF
--- a/pcweb/pages/docs/component.py
+++ b/pcweb/pages/docs/component.py
@@ -92,11 +92,6 @@ class Source(Base):
                 # This isn't a prop, so continue.
                 continue
 
-            # redundant check just to double-check line above prop is a comment
-            assert (
-                self.code[i - 1].strip().startswith("#")
-            ), f"Expected comment, got {comment}"
-
             # Get the comment for this prop.
             comment = Source.get_comment(comments)
             # reset comments


Alek99 (approved):

milochen0418:
We should not delete this code on pcweb
```python
            # redundant check just to double-check line above prop is a comment
            assert (
                self.code[i - 1].strip().startswith("#")
            ), f"Expected comment, got {comment}"
```

It's the comment on the field in pynecone's component.
So fixing the source code can solve this issue. 
https://github.com/pynecone-io/pynecone/pull/1081/files

When we fix it, the comment part of the field becomes the document for this field. 
Screenshot 2023-05-26 at 4 27 00 PM

In the component.py, the class Source parses the source code of a component. 
That's how it works, which is also why we must keep these lines.

milochen0418:
We should not delete this code on pcweb
```python
            # redundant check just to double-check line above prop is a comment
            assert (
                self.code[i - 1].strip().startswith("#")
            ), f"Expected comment, got {comment}"
```

It's the comment on the field in pynecone's component.
So fixing the source code can solve this issue. 
So the following PR cane solve this issue. 
https://github.com/pynecone-io/pynecone/pull/1081/files

When we fix it, the comment part of the field becomes the document for this field. 
Screenshot 2023-05-26 at 4 27 00 PM

In the component.py, the class Source parses the source code of a component. 
That's how it works, which is also why we must keep these lines.

PR #91, opened 2023-06-01
add doc for local fonts in custom stylesheet

Lendemor:


picklelo (approved):
sweet I didn't know we could do this :)

PR #92, opened 2023-06-01
Lendemor/fix middleware page

Lendemor:
Temporary fix to resolve the conflict between `advanced-guide/middleware` and `api-reference/middleware`

Change the former to `advanced-guide/use-middleware` instead.

Can be rolled back once we find the proper fix to avoid name conflict when building pages.


Lendemor:
(also include some formatting fix by running black on project)

picklelo (approved):
I think this may affect our urls - as we use the function name for the url. May have to rethink the current `docpage` as we keep adding pages as there will be more conflicts like this I'm guessing.

PR #93, opened 2023-06-01
add reserved routes in api_route page

Lendemor:
Add explanation in `api-route` for the 3 reserved endpoints : `event`, `upload` and `ping`

### DIFF
--- a/pcweb/pages/docs/advanced_guide/api_routes.py
+++ b/pcweb/pages/docs/advanced_guide/api_routes.py
@@ -43,4 +43,33 @@ def api_routes():
         doctext(
             "This is useful for creating a backend API that can be used for purposes other than your Pynecone app. ",
         ),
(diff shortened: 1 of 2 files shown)


pcweb/pages/docs/advanced_guide/api_routes.py
```diff
@@ -43,4 +43,33 @@ def api_routes():
...
+        doctext(
+            "The expected return is ",
+            pc.code('"pong"'),
+        ),
+        subheader("Event", level=2),
+        doctext(
+            pc.code("localhost:8000/event"),
+            ": the frontend will use this route to notify the backend that an event occurred",
+        ),
+        pc.alert(
```
picklelo, line 64: we can use the `docalert` component for consistent styling (can do in a follow up)

picklelo (approved):
nice!

PR #95, opened 2023-06-02
doc for using async event handler, calling them, and yielding

Lendemor:


pcweb/pages/docs/state/events.py
```diff
@@ -80,67 +80,115 @@ def alert(self):
...
+            await asyncio.sleep(0.2)
 
+            if self.count % 2 == 0:
+                # If the number is even, divide by 2.
+                self.count /= 2
+            else:
+                # If the number is odd, multiply by 3 and add 1.
+                self.count = self.count * 3 + 1
+            yield
+        yield
```
picklelo, line 104: I don't think we need this last `yield`

pcweb/pages/docs/state/events.py
```diff
@@ -223,6 +265,20 @@ def events():
         doctext(
             "Setters are a great way to make your code more concise. But if you want to do something more complicated, you can always write your own function in the state."
         ),
+        subheader("Yielding multiple update"),
```
picklelo, line 268: Can we use title case here - `Yielding Multiple Updates`

pcweb/pages/docs/state/events.py
```diff
@@ -223,6 +265,20 @@ def events():
...
+        doctext(
+            "A regular event handler will send a StateUpdate when it has finished running.",
```
picklelo, line 270: Let's either do `pc.code(StateUpdate)` or call it "a state update".

Also - we need to add a space after each sentence or they will be joined together.

pcweb/pages/docs/state/events.py
```diff
@@ -240,11 +296,20 @@ def events():
...
-            "However, events can also be triggered by event handlers. ",
-            "To do this, you can return either an event, or an event chain from an event handler.",
+            "However, an event handler can also trigger others event handlers.",
+        ),
+        doctext("This way to call event handlers work with either async or non-async events."),
+        doctext(
+            pc.alert(
+                icon=True,
+                title=pc.text(
+                    "Be sure to use the class name State (or any substate) rather than ", pc.code("self"), "."
```
picklelo, line 306: Let's do `pc.code(State)`

picklelo (dismissed):
Just some small style comments

picklelo (approved):
awesome!

PR #113, opened 2023-06-23
Add autogenerated docs feature with Markdown parser

kbrgl:
This adds a docs/ directory that sets the base for autogenerating docs or a blog. Any Markdown files under docs/docs/pages/ will be directly mapped to a route (e.g. docs/docs/pages/index.md will map to /).

Further, the docs/ directory includes a parser in `utils.py` that parses out code blocks annotated with ```reflex, evaluates them, and interpolates the resulting Reflex component in the output. Thus, to put a blue heading, you can write the following Markdown:

~~~markdown
```reflex
pc.heading("Hello, world!", color="blue")
```

Other **Markdown** here.
~~~


picklelo (approved):
nice this is a good start, we can use this for the blog page next

PR #121, opened 2023-06-29
add browser docs (cookie & local storage)

Lendemor:


masenf (commented):
might be worth mentioning that `get_local_storage()` cannot be used for direct rendering

masenf (approved):

PR #128, opened 2023-07-07
update self_hosting page and run black on project

Lendemor:


pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -9,20 +9,42 @@
...
+├── rxconfig.py
+├── Dockerfile
+└── requirements.txt"""
+
+docker_build_example = """$ docker build -t reflex-project:latest ."""
+docker_run_example = (
+    """$ docker run -d -p 3000:3000 -p 8000:8000 --name reflex reflex:latest"""
+)
+
+lighthouse_ref_link = "https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk"
```
masenf, line 42: i would use https://developer.chrome.com/docs/lighthouse/overview/ instead, since the chrome plugin is actually deprecated and lighthouse is built into devtools now

pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -119,49 +133,43 @@ def self_hosting():
...
+        subheader("Optimization"),
         doctext(
-            "After all changes have been made, the container image can now be created as follows.",
+            "When self-hosting a Reflex app, you might encounter the need to improve its performance. ",
         ),
-        doccode(
-            """$ docker build -t reflex-project:latest .""",
-            language="bash",
+        doctext(
+            "Once you have a public URL leading to your app, you can use tools like ",
```
masenf, line 151: i don't think a public url is necessary for lighthouse benchmarking... but the hosting mechanism may influence the score.

masenf (commented):
side comment: keeping the content changes and the formatting changes in separate commits makes reviewing the patch easier.

picklelo (commented):
Looks good! Needs to be rebased now but will merge in after

pcweb/pages/docs/hosting/self_hosting.py
Lendemor: it was necessary with the plugin, seems like the devtools integration is better.

pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -119,49 +130,59 @@ def self_hosting():
...
+                " is very slow and will result in bad performance.",
+            ),
+            status="warning",
         ),
+        subheader("Compression", level=1),
         doctext(
-            "Finally, you can start your Reflex container service as follows.",
+            "Among others, enabling compression is one of the foremost point to optimize. \n",
+            "We use Next.js to run the frontend, and while it has gzip compression enabled by default, ",
+            "it is usually recommanded to offload it to your reverse proxy (ie. Nginx or Apache)",
```
masenf, line 172:
```suggestion
            "it is usually recommended to offload it to your reverse proxy (ie. Nginx or Apache)",
```

pcweb/pages/docs/hosting/self_hosting.py
```diff
@@ -7,22 +7,43 @@
...
+├── assets
+├── hello
+│   ├── __init__.py
+│   └── hello.py
+├── rxconfig.py
+├── Dockerfile
+└── requirements.txt"""
+
+docker_build_example = """$ docker build -t reflex-project:latest ."""
+docker_run_example = """$ docker run -d -p 3000:3000 -p 8000:8000 --name reflex reflex:latest"""
```
masenf, line 41:
```suggestion
docker_run_example = """$ docker run -d -p 3000:3000 -p 8000:8000 --name reflex reflex-project:latest"""
```

the image to run should be the same tag as the image built in the previous step

PR #151, opened 2023-07-31
self_hosting: explain more about api_url and ports

masenf:
yeah, this conflicts with #128 not sure what to do about that at this point...


Alek99 (approved):
nice, lgtm!

PR #153, opened 2023-07-31
Get rid of value

Alek99:
This should allow the integration test to pass in the main repo after we stopped supporting value for input


masenf (commented):
i have rewritten this section of the doc. let me push it up

Alek99:
Ok perfect will close this

masenf:
Follow up in #154

PR #156, opened 2023-08-01
Added preview images and removed loading image in state

Alek99:


pcweb/pcweb.py
```diff
@@ -23,7 +23,7 @@
         route.path,
         "Reflex",
         description="Performant, customizable web apps in pure Python. Deploy in seconds.",
-        image="preview.png",
+        image="/index_preview.png",
```
masenf:
```suggestion
        image="previews/index_preview.png",
```
This one is nlot like the others

masenf (approved):

pcweb/pcweb.py
Alek99: Fixed in my other pushes it think

picklelo (approved):
images look nice

PR #168, opened 2023-08-03
wrapping_react: add section on Local Components

masenf:
image

https://github.com/reflex-dev/reflex/issues/1195


picklelo (approved):
This is awesome! Maybe we should make a separate directory for custom components? Or you think its okay in the assets

pratyushseq:
@masenf How do we use npm packages in these local components? For example if I want to use react-datepicker in the above-mentioned Welcome component, how do I install and use it?

PR #173, opened 2023-08-08
Document next/script component

masenf:
# New warning on Html page

image

# All New `rx.script` docs and examples

image
image
image
image

Not sure why the docstring is rendering with monospace text like that 🤔 leading 4 spaces maybe interpreted by markdown as a code block?

```python
class Script(Component):
    """Next.js script component.

    Note that this component differs from reflex.components.base.document.NextScript
    in that it is intended for use with custom and user-defined scripts.

    It also differs from reflex.components.base.link.ScriptTag, which is the plain
    HTML <script> tag which does not work when rendering a component.
    """

    library = "next/script"
    tag = "Script"
    is_default = True
```

### DIFF
--- a/pcweb/component_list.py
+++ b/pcweb/component_list.py
@@ -189,7 +189,7 @@ graphing_list = [
 ]
 
 # Other
-other_list = [[rx.Html]]
+other_list = [[rx.Html], [rx.Script]]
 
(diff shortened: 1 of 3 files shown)


Alek99 (approved):
lgtm

PR #179, opened 2023-08-18
The property api_key is updated.

mouredev:
In the tutorial, the property name api_key from OpenAI has been used in uppercase. The property only exists in lowercase.


masenf (approved):
Good catch. Thanks for your contribution.

PR #181, opened 2023-08-21
Update chat tutorial and Input/TextArea/Editable docs

martinxu9:
## Summary 
- Update chat tutorial and Input/TextArea/Editable docs.


pcweb/pages/docs/tutorial/adding_state.py
```diff
@@ -236,34 +239,34 @@ def adding_state():
...
-        #     "We use the ",
-        #     rx.code("set_value"),
-        #     " function to set the value of the ",
-        #     rx.code("question"),
-        #     " state var to an empty string. ",
-        #     "See the ",
-        #     doclink("special events docs", special_events.path),
-        #     " for more info on special events. ",
-        # ),
+        subheader("Clearing the Input"),
```
masenf, line 242: this section doesn't match the code anymore

pcweb/pages/docs/component_lib/forms/input.py
```diff
@@ -126,15 +123,20 @@ def render_input():
...
+            "still typing. ",
             "This allows a state var to directly control the ",
             rx.code("value"),
             " prop from the backend without the user experiencing input lag. ",
+            "For advanced use cases, you can tune the debounce delay ",
+            "by setting the ",
+            rx.code("debounce_timeout"),
+            " prop. ",
+            "You can find examples of how it is used in ",
+            doclink("DebouncedInput", href="/docs/library/forms/debounceinput"),
```
masenf, line 138:
```suggestion
            doclink("DebounceInput", href=render_debounceinput.path),
```

masenf (changes requested):

pcweb/pages/docs/tutorial/adding_state.py
martinxu9: huh, it's not ready for review yet actually, let me mark it as draft

pcweb/pages/docs/component_lib/forms/input.py
martinxu9: This is what I tried first, but this doesn't seem to work. There's those returning a page or something, those have the path attribute

masenf: ah, it looks like `render_debounceinput` might be missing the `@docpage` decorator? not sure exactly.

martinxu9: adding the decorator caused some other errors for me, didn't look different from other existing examples with the decorator, strange.

pcweb/pages/docs/tutorial/adding_state.py
```diff
@@ -236,34 +285,28 @@ def adding_state():
...
+        doctext(
+            "Currently the input doesn't clear after the user clicks the button. ",
+            "We can fix this by binding the value of the input to ",
+            rx.code("question"),
+            " and creating an event ",
```
picklelo, line 293: should be "creating an event handler"

pcweb/pages/docs/tutorial/adding_state.py
```diff
@@ -236,34 +285,28 @@ def adding_state():
...
+            rx.code("clear_question"),
+            " to clear it. ",
+        ),
+        docdemobox(code_out2),
+        doccode(code2),
+        doccode(state2),
+        doctext(
+            "After setting the question, we return an event from our ",
```
picklelo, line 301: I think we should modify this to
```python
def answer(self):
    # Our chatbot is not very smart right now...
    answer = "I don't know!"
    self.chat_history.append((self.question, answer))
    self.question = ""
```
Then we can get rid of the second event handler. I had it that way in the commented code because this method wasn't working at the time.

pcweb/pages/docs/tutorial/adding_state.py
```diff
@@ -236,34 +285,28 @@ def adding_state():
...
+            rx.code("answer"),
+            " event handler to clear the input. ",
+            "Learn more about returning events in the ",
```
picklelo, line 304: Then we can remove this section here also.

pcweb/pages/docs/tutorial/final_app.py
```diff
@@ -32,17 +32,31 @@ def answer(self):
     # Add to the answer as the chatbot responds.
     answer = ""
     self.chat_history.append((self.question, answer))
+    # After adding the question into the chat history,
+    # we clear the question and yield such that
+    # the frontend updates now instead of sending an event.
+    self.clear_question()
```
picklelo, line 38: This can just be `self.question = ""` after we remove the other one

pcweb/pages/docs/tutorial/final_app.py
```diff
@@ -32,17 +32,31 @@ def answer(self):
     # Add to the answer as the chatbot responds.
     answer = ""
     self.chat_history.append((self.question, answer))
+    # After adding the question into the chat history,
+    # we clear the question and yield such that
+    # the frontend updates now instead of sending an event.
```
picklelo, line 37: I think we can just add one comment here saying `# Clear the question input` and then right above the `yield` add the other comment saying `# Yield here to clear the frontend input before continuing`

(Realized this comment should apply to the previous page also)

pcweb/pages/docs/tutorial/final_app.py
```diff
@@ -32,17 +32,31 @@ def answer(self):
...
+    self.clear_question()
+    yield
```
picklelo, line 39: nit: can we add a blank line after this to  separate out the parts of the code

picklelo (commented):
Great work, just some minor comments

martinxu9:
@picklelo @masenf updated based on comments, PTAL

pcweb/pages/docs/tutorial/adding_state.py
```diff
@@ -83,18 +93,27 @@ def answer(self):
...
         )
     )
+
+
 def action_bar1() -> rx.Component:
     return rx.hstack(
-        rx.input(placeholder="Ask a question", on_blur=ChatappState.set_question, style=style.input_style),
+        rx.input(
+            placeholder="Ask a question",
+            on_change=ChatappState.set_question,
```
masenf, line 110:
```suggestion
            value=ChatappState.question
            on_change=ChatappState.set_question,
```

does this also need the `value` prop set, so that it can be cleared?

if so, check for other instances of input with `on_change` but no `value` as I saw at least one other

masenf (changes requested):

picklelo (approved):
looks good to me!

martinxu9:
@masenf guess needs your approval as well since you requested changes

masenf (approved):

PR #191, opened 2023-08-29
Include `rx.clear_selected_files` in the example

martinxu9:
## Summary
- Include `rx.clear_selected_files` in the example.

### DIFF
--- a/pcweb/pages/docs/component_lib/forms/forms.py
+++ b/pcweb/pages/docs/component_lib/forms/forms.py
@@ -237,7 +237,7 @@ def render_upload():
             rx.code("rx.selected_files"),
             " special Var. To clear the selected files, you can use another special Var ",
             rx.code("rx.clear_selected_files"),
-            ". ",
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/component_lib/forms/forms.py
```diff
@@ -231,7 +235,9 @@ def render_upload():
             "Selecting a file will add it to the browser's file list, which can be rendered ",
             "on the frontend using the ",
             rx.code("rx.selected_files"),
-            " special Var. "
+            " special Var. To clear the selected files, you can use another special Var ",
```
masenf, line 238:
```suggestion
            " special Var. To clear the selected files, you can use another special Var as an event handler ",
```

masenf (approved):

pcweb/pages/docs/component_lib/forms/forms.py
martinxu9: :+1

PR #192, opened 2023-09-06
State / Vars: add section+example on using Cookie and LocalStorage

masenf:
image
Screenshot 2023-09-06 at 1 28 08 PM


Alek99 (approved):
Works for me, really clean api as well great job with this!

PR #199, opened 2023-09-11
custom_stylesheets: show examples with absolute paths

masenf:
Relative paths are not working, since custom stylesheets get included in every page and not ever page is in the "root" of the site.

https://github.com/reflex-dev/reflex/issues/1740


picklelo:
@masenf Looks like this change still never made it in 😅 - can we update this to the `.md` file and merge it

masenf:
added it to tomorrow's queue

picklelo (approved):

PR #200, opened 2023-09-13
Added warning for macOS users using Apple Silicon

divin:
Thanks to the help of @masenf, this [issue](https://github.com/reflex-dev/reflex/issues/1803) has been resolved. The new package manager `fnm` seems to require `Rosetta 2` to work. As this may be crucial for developers using Apple silicon chips, I think it makes sense to add this information to the installation guide. In case other users encounter the same problem.

You can see the changes in the image below, let me know if you don't like the wording or styling. I tried the `doccode` component, but it does not look good inside the alert component. So I used the regular `rx.code` component.

Screenshot 2023-09-13 at 06 39 27

### DIFF
--- a/pcweb/pages/docs/getting_started/installation.py
+++ b/pcweb/pages/docs/getting_started/installation.py
@@ -27,6 +27,12 @@ def installation():
             " for optimal performance.",
             status="warning",
         ),
+        docalert(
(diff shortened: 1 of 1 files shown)


picklelo (approved):
Thanks for adding - we're looking into removing FNM in an upcoming release now that Bun supports NextJS, so hopefully we can remove the warning then.

PR #201, opened 2023-09-13
fixed rx.upload sample code in /docs/library/forms/upload by changing…

Anga205:
the sample code for rx.uploads in /docs/library/forms/upload defines rx.uploads img instide state but calls State.imgs which throws an error, which can be fixed by changing it to State.img which is what i have done in this edit


picklelo (approved):
nice catch!

PR #205, opened 2023-09-20
docs for disabling tailwind css

ElijahAhianyo:
Screenshot 2023-09-20 at 8 05 07 PM


masenf (commented):
I would avoid the extra stuff in the Config so that the `tailwind=None` is more obvious

ElijahAhianyo:
> I would avoid the extra stuff in the Config so that the `tailwind=None` is more obvious

Got it. Fixed

masenf (approved):

PR #207, opened 2023-09-20
Add initial flexdown docs

picklelo:


pcweb/flexdown.py
```diff
@@ -0,0 +1,147 @@
...
+        if line.startswith("```reflex"):
+            line = line[len("```reflex") :]
+            in_reflex_block = True
+        elif in_reflex_block and line.startswith("```"):
+            # End reflex block.
+            in_reflex_block = False
+            try:
+                result = eval("\n".join(current_block))
+            except Exception as e:
+                print(f"Error in reflex block: {str(e)}")
```
masenf, line 123: File and line number would be extremely helpful here

```python
raise FlexdownParseError(lineno=lineno) from e
```

Then down below in `read`, you can catch the exception and print a helpful error message with the filename, line, and original exception.

picklelo: Will add this in a follow-up, created this ticket: https://github.com/reflex-dev/reflex-web/issues/208

pcweb/flexdown.py
```diff
@@ -0,0 +1,147 @@
...
+    front_matter, source = parse_markdown_front_matter(source)
+    if isinstance(front_matter, str):
+        exec(front_matter)
+
+    lines = source.split("\n")
+    output = []
+    in_reflex_block = False
+    current_block = []
+
+    for line in lines:
```
tgberkeley, line 107: This appears to cause our python code in markdown to break if we have vertical spaces in the code:

```python
import reflex as rx


config = rx.Config(
    app_name="hello",
)
```
Here the code is not all in the code box, it stops at config = rx.Config( app_name="hello", )

picklelo: Got it, will add this also in my follow up.

tgberkeley (approved):

PR #212, opened 2023-09-22
update the signup form to automatically send the email to loops via API

tgberkeley:


pcweb/pages/index.py
```diff
@@ -54,6 +57,27 @@ def close_c2a(self):
...
+        }
+
+        try:
+            with httpx.Client() as client:
+                response = client.post(url, headers=headers, json=contact_data)
+                response.raise_for_status()  # Raise an exception for HTTP errors (4xx and 5xx)
+ 
+        except httpx.RequestError as e:
+            print(f"An error occurred: {e}")
+            return None
```
picklelo, line 78: don't need this return

picklelo (approved):

PR #220, opened 2023-09-27
adding docs for audio and video components and updating docs so that they take in parent props as well for components

tgberkeley:
adding docs for audio and video components

### DIFF
--- a/docs/library/media/audio.md
+++ b/docs/library/media/audio.md
@@ -9,10 +9,10 @@ audio_example = """rx.audio(
 )"""
 
 ---
-The audio component can display an audio given an src path as an argment. This could either be a local path from the assets folder or an external link.
+The audio component can display an audio given an src path as an argument. This could either be a local path from the assets folder or an external link.
 
 ```reflex
 docdemo(audio_example)
 ```
 
-If we had a local file in the `assets` folder named `test.mp3` we could set `url=test.mp3` to view the audio file.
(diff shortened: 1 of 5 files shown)


docs/library/media/audio.md
```diff
@@ -0,0 +1,18 @@
...
+from pcweb.templates.docpage import docdemo
+
+audio_example = """rx.audio(
+    url="https://www.learningcontainer.com/wp-content/uploads/2020/02/Kalimba.mp3", 
+    width="400px",
+    height="auto"
+)"""
+
+---
+The audio component can display an audio given an src path as an argment. This could either be a local path from the assets folder or an external link.
```
masenf, line 12:
```suggestion
The audio component can display an audio given an src path as an argument. This could either be a local path from the assets folder or an external link.
```

docs/library/media/audio.md
```diff
@@ -0,0 +1,18 @@
...
+
+```reflex
+docdemo(audio_example)
+```
+
+If we had a local file in the `assets` folder named `test.mp3` we could set `url=test.mp3` to view the audio file.
```
masenf, line 18:
```suggestion
If we had a local file in the `assets` folder named `test.mp3` we could set `url="/test.mp3"` to view the audio file.
```

Better to show example with leading slash, so that pages other than `/` (index) can access the file.

masenf (commented):
same comments here apply to the video component too.

picklelo (approved):

PR #225, opened 2023-09-29
Button docs

ElijahAhianyo:
Moved the button and button group docs to flex down with improvements


tgberkeley:
We must also go into the component_lists.py file (pcweb/component_list.py) and change from all 3 button items being in one list to having them in 2 separate lists ([rx.Button, rx.IconButton],
    [rx.ButtonGroup],) so that the button group page renders

masenf (approved):

PR #226, opened 2023-09-29
[WIP]Special events docs

ElijahAhianyo:


docs/api-reference/browser_storage.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+
+## rx.LocalStorage
+Represents a state Var that is stored in localStorage in the browser.
+
+Parameters
+- `encoding`: The encoding to use
+- `errors`: The error handling scheme to use.
+- `name`: THe name of the storage key on the client side.
```
tgberkeley, line 63: Accidental capital H in the word the here

tgberkeley:
It might be useful to add in this section http://reflex.dev/docs/state/vars#client-side-storage, examples of where we use cookies and where we use local storage, i.e. what are the different use cases and where to use each

docs/api-reference/browser_storage.md
ElijahAhianyo: Fixed

ElijahAhianyo:
> It might be useful to add in this section http://reflex.dev/docs/state/vars#client-side-storage, examples of where we use cookies and where we use local storage, i.e. what are the different use cases and where to use each

Since this is under the API reference section, the plan was to not have it inundated with so much content other than information on how what the API looks like and what it does firsthand. Under the Var section (https://reflex.dev/docs/state/vars/#client-side-storage) however, is explained what the supported client storages are. Perhaps a detailed explanation with use cases as you've mentioned could go there instead

tgberkeley:
We should have all the code blocks also rendering the code so we can see the output components and let people play around with them on their machines. We should use docdemo in the code instead of doccode. Also we have a new way of creating flex down code, where we create it as code now instead of as strings.

tgberkeley:
Here is an example from our state docs:
```python exec
def qa(question: str, answer: str) -> rx.Component:
    return rx.box(
        rx.box(rx.text(question, style=style.question_style), text_align="right"),
        rx.box(rx.text(answer, style=style.answer_style), text_align="left"),
        margin_y="1em",
        width="100%",
    )


def chat1() -> rx.Component:
    return rx.box(
        rx.foreach(
            ChatappState.chat_history, lambda messages: qa(messages[0], messages[1])
        )
    )


def action_bar1() -> rx.Component:
    return rx.hstack(
        rx.input(
            placeholder="Ask a question",
            on_change=ChatappState.set_question,
            style=style.input_style,
        ),
        rx.button("Ask", on_click=ChatappState.answer, style=style.button_style),
    )


def rendered_code_binding_state_to_components():
    return rx.container(
        chat1(),
        action_bar1(),
    )
```



```python exec

def show_code_binding_state_to_components():

    # chatapp.py
    from chatapp.state import State
    ...

    def chat() -> rx.Component:
        return rx.box(
            rx.foreach(
                State.chat_history,
                lambda messages: qa(messages[0], messages[1])
            )
        )

    ...

    def action_bar() -> rx.Component:
        return rx.hstack(
            rx.input(placeholder="Ask a question", on_change=State.set_question, style=style.input_style),
            rx.button("Ask", on_click=State.answer, style=style.button_style),
        )

```


```python eval
docdemo(
    inspect.getsource(show_code_binding_state_to_components).replace("def show_code_binding_state_to_components():", ""),
    comp=rendered_code_binding_state_to_components()
)

tgberkeley (approved):

PR #230, opened 2023-10-01
fix tutorial State variable naming error

imyhui:
fix https://github.com/reflex-dev/reflex/issues/1735


picklelo (approved):
thanks for fixing!

PR #243, opened 2023-10-06
Change recharts Namespace to rx.recharts.<component>

ElijahAhianyo:


masenf (commented):
we've got some missing spots in the text as well, i can push another update.

masenf (approved):
reviewed all pages for graph components and it looks good to me!

PR #254, opened 2023-10-09
Add links to all header tags

picklelo:


pcweb/templates/docpage.py
```diff
@@ -398,38 +374,28 @@ def docheader2(
...
     return rx.box(
-        *children,
-        id=id_,
+        rx.hstack(
+            rx.heading(text, _as=tag, id=id_, style=style, **props),
+            rx.icon(
+                tag="link",
+                style=icon_style,
+                color=styles.ACCENT_COLOR,
+                on_click=lambda: rx.set_clipboard(State.current_page + "#" + id_)
```
masenf, line 395: i think we should make this a real link and set the `href` too, as many people (myself included) right click these guys and "Copy Link"

masenf (changes requested):

pcweb/templates/docpage.py
picklelo: Good call, updated to use a link instead

Alek99 (approved):
lgtm

masenf (approved):
The thought the "click to copy" thing was cool too, was it not working to have both?

picklelo:
> The thought the "click to copy" thing was cool too, was it not working to have both?

I brought this back now, so it should do both.

masenf (approved):

PR #263, opened 2023-10-11
Graphing blog

Alek99:


tgberkeley:
Maybe we could add in another simple non-interactive example

blog/2023-10-11-graphing-update.md
```diff
@@ -0,0 +1,301 @@
...
+            self.stream = True
+        while self.stream:
+            for i in range(len(self.data)):
+                async with self:
+                    self.data[i]["uv"] = random.randint(0, 100)
+                    self.data[i]["pv"] = random.randint(100, 200)
+                    self.data[i]["amt"] = random.randint(200, 300)
+            await asyncio.sleep(3)
+```
+
```
picklelo, line 75: typo: reference

blog/2023-10-11-graphing-update.md
```diff
@@ -0,0 +1,301 @@
...
+    \{"name": "E", "uv": 20, "pv": 140, "amt": 230\},
+    \{"name": "F", "uv": 40, "pv": 170, "amt": 250\},
+    \{"name": "G", "uv": 50, "pv": 190, "amt": 260\},
+]
+```
+
+Next we define a `StreamingState` class that will be used to store the data and update it with an event handler:
+
+```python
+class StreamingState(State):
```
picklelo, line 56: This should inherit from `rx.State`

Also we should use `inspect.getsource` to make sure the code matches the actual sourc

blog/2023-10-11-graphing-update.md
```diff
@@ -0,0 +1,301 @@
...
+    data = data
```
picklelo, line 57: Can we add some comments for these vars. And also maybe add type annotations to encourage users to use them

blog/2023-10-11-graphing-update.md
```diff
@@ -0,0 +1,301 @@
...
+                    self.data[i]["uv"] = random.randint(0, 100)
+                    self.data[i]["pv"] = random.randint(100, 200)
+                    self.data[i]["amt"] = random.randint(200, 300)
+            await asyncio.sleep(3)
+```
+
+The result is a live updating graph that looks like this:
+
+```python eval
+rx.vstack(
```
picklelo, line 112: Same here we should just do
```python eval
return index()
```
rather than rewriting the code

picklelo (commented):
Looks good to me

picklelo (approved):

PR #271, opened 2023-10-17
Optimize graphing blog

picklelo:
Before we were yielding 3x for each update, causing the graph animation to lag.


masenf (approved):
good find

PR #275, opened 2023-10-20
expand database table and query sections

masenf:
Preview? https://pcweb-gold-ring.dev.reflexcorp.run/docs/database/overview/


masenf:
reference where the rx.Model `id` field comes from..

masenf:
add more explicit description of `session.commit()` (and does session.rollback() exist?)

ElijahAhianyo:
A little note of connecting to the DB, using `DBConfig` object is deprecated (not sure if this right wording since it throws a `ValueError` when used). We may want to take it out of the docs so users are not encouraged to try that.

Alek99 (approved):

PR #276, opened 2023-10-23
Wrapping React Guide Improvements

Alek99:


masenf (commented):
I think only the opening curly brace needs to be escaped... that said, we should consider changing the inline eval syntax to use `{{` and `}}` like jinja because those appear so much less frequently in code samples.

docs/wrapping-react/logic.md
```diff
@@ -0,0 +1,97 @@
...
+
+except ImportError:
+    pass
+```
+
+## Event Triggers
+
+As seen in our [events section](https://reflex.dev/docs/state/events/), we can use event triggers to handle events in our Reflex apps. When wrapping your own react components, you can use the `get_event_triggers` method to define event triggers for your component.
+
+Sometimes these event trigger may take in arguments, for example, the `on_change` event trigger for the `HexColorPicker` component we saw in the [wrapping react section](https://reflex.dev/docs/wrapping-react/wrapping-react/). In this case, we can use a lambda function to pass in the event argument to the event trigger.
```
masenf, line 84:
```suggestion
Sometimes these event trigger may take in arguments, for example, the `on_change` event trigger for the `HexColorPicker` component we saw in the [wrapping react section](https://reflex.dev/docs/wrapping-react/wrapping-react/). In this case, we can use a lambda function to pass in the event argument to the event trigger. The function associated with a trigger maps args for the javascript trigger to args that will be passed to the backend event handler function.
```

docs/wrapping-react/wrapping-react.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Spline Example
+
+
+Let's start with a library called [Spline](https://spline.design/). Spline is a tool for creating 3D scenes and animations. It's a great tool for creating interactive 3D visualizations.
+
+
+We have some code that creates a Spline scene. We want to wrap this code in a Reflex component so that we can use it in our Reflex app.
+
+```javascript
```
tgberkeley, line 26: make a comment in the text that this is javascript perhaps

docs/wrapping-react/wrapping-react.md
```diff
@@ -0,0 +1,170 @@
...
+    is_default = True
+
+    lib_dependencies: list[str] = ["@splinetool/runtime"]
+```
+
+
+Here the library is `@splinetool/react-spline` and the tag is `Spline`. In the next section we will go into a deep dive on imports but we also set `is_default = True` because the tag is the default export from the module.
+
+
+Additionally, we can specify any props that the component takes. In this case, the `Spline` component takes a `scene` prop, which is the URL of the Spline scene.
```
tgberkeley, line 54: where can I find the props that the component takes?

docs/wrapping-react/wrapping-react.md
```diff
@@ -0,0 +1,170 @@
...
+    )
+```
+
+## ColorPicker Example
+
+Similar to the Spline example we start with defining the library and tag. In this case the library is `react-colorful` and the tag is `HexColorPicker`.
+
+We also have a var `color` which is the current color of the color picker.
+
+Since this componnent has interaction we must specify any event triggers that the component takes. The color picker has a single trigger `on_change` to specify when the color changes. This trigger takes in a single argument `color` which is the new color.
```
tgberkeley, line 101: component here is spelt wrong with 2 n's

docs/wrapping-react/wrapping-react.md
```diff
@@ -0,0 +1,170 @@
...
+
+```python exec
+class ColorPicker(rx.Component):
+    library = "react-colorful"
+    tag = "HexColorPicker"
+    color: rx.Var[str]
+
+    def get_event_triggers(self) -> dict[str, Any]:
+        return {
+            **super().get_event_triggers(),
```
tgberkeley, line 112: no explanation as to where the **super comes from or why it is needed?

docs/wrapping-react/wrapping-react.md
picklelo: +1

docs/wrapping-react/wrapping-react.md
```diff
@@ -0,0 +1,170 @@
...
+
+export default function App() {
+  return (
+    <Spline scene="https://prod.spline.design/up1SQcRLq1s6yks3/scene.splinecode" />
+  );
+}
+```
+
+The two most important props are `library`, which is the name of the npm package, and `tag`, which is the name of the React component.
+
```
picklelo, line 37: Agreed with Tom's above comment - and then we can say something like "And below is the Reflex code to wrap it.."

docs/wrapping-react/wrapping-react.md
```diff
@@ -0,0 +1,170 @@
...
+```python
+class Spline(rx.Component):
+    """Spline component."""
+
+    library = "@splinetool/react-spline"
+    tag = "Spline"
+    scene: Var[str] = "https://prod.spline.design/Br2ec3WwuRGxEuij/scene.splinecode"
+    is_default = True
```
picklelo, line 45: Maybe move the `is_default and `lib_dependencies` tags above `scene` so that the library props are together, and the "real" props are separated.

picklelo (approved):
great work

PR #279, opened 2023-10-25
Foreach docs fix

ElijahAhianyo:
There's a bug where the first Foreach example in the docs shows the color index instead of the names. This is because The second example (`ForeachIndexState`) also defines a `colored_box` function with a different signature to that of the first example(`ForeachState`). Since the exec blocks are executed before rendered, the second `colored_box` function overrides the first. This PR renames the second `colored_box` to fix the conflict


picklelo (approved):
nice catch!

PR #285, opened 2023-10-27
add page on Browser Javascript interface

masenf:
REF-809

Sort of goes with wrapping react, but is more about implementing custom functionality and interfacing it with State


Alek99 (approved):
This is so cool, all  love the dust storm!

Really clear explanations to nice

PR #304, opened 2023-11-10
data table tutorial

tgberkeley:


docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
+```python exec
+import reflex as rx
+from pcweb.templates.docpage import docdemobox, doclink
+from pcweb.base_state import State
+from pcweb.pages.docs.datatable_tutorial.states import DataTableState, DataTableState2
+```
+
+
+
+# Data Table Tutorial
```
masenf, line 10: If this isn't for the `rx.data_table` component, maybe the heading should be "Data Editor Tutorial"?

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+
+```python eval
+rx.alert(
+        rx.alert_icon(),
+        rx.alert_title(f"There is another ",
+        doclink("datatable component", href="https://reflex.dev/docs/library/datadisplay/datatable/"),
+        ", but this is only used for displaying data and there is no possible user interaction while using this component"),
```
masenf, line 17:
```suggestion
        ", which is only used for displaying data and does not support user interactivity or editing"),
```

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+        status="info",
+    )
+```
+
+
+```python eval
+rx.box(height="2em")
+```
+We need to start by defining our `columns` that will then take in our data. These columns should be a list of dict, where each dict describes the associated columns.
```
masenf, line 26:
```suggestion
We need to start by defining our columns that describe the shape of our data. The column var should be typed as a `list` of `dict` (`list[dict]`), where each item describes the attributes of a single column in the table. 
```

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+    )
+```
+
+
+```python eval
+rx.box(height="2em")
+```
+We need to start by defining our `columns` that will then take in our data. These columns should be a list of dict, where each dict describes the associated columns. 
+
+Each dict has these variables below:
```
masenf, line 28:
```suggestion
Each column dict recognizes the keys below:
```

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+1. title: The text to display in the header of the column
+2. id: An id for the column, if not defined, will default to a lower case of title
+3. width: The width of the column
+4. type: The type of the columns, default to "str"
```
masenf, line 32: Use code marks on the key names

```suggestion
1. `title`: The text to display in the header of the column
2. `id`: An id for the column, if not defined, will default to a lower case of title
3. `width`: The width of the column (in pixels?)
4. `type`: The type of the columns, default to "str"
```

For `type`, how can i find out what types are available?

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+Below we define a `DataTableState` which has Harry Potter data. Here an example of a dict in the cols var is: {
+            "title": "Name",
+            "type": "str",
+            "width": 300,
+        }.
```
masenf, line 39:
```suggestion
Below we define `DataTableState` with columns definitions in the `cols` var, and data about Harry Potter characters in the `data` var..
```

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+        ["4", "Albus Percival Wulfric Brian Dumbledore", "Late August 1881", True, "Gryffindor", "15' Elder Thestral tail hair core", "Phoenix", "Half-blood"],	
+        ["5", "Rubeus Hagrid", "6 December 1928", False, "Gryffindor", "16'  Oak unknown core", "None", "Part-Human (Half-giant)"], 
+        ["6", "Fred Weasley", "1 April, 1978", True, "Gryffindor", "Unknown", "Unknown", "Pure-blood"], 
+        ["7", "George Weasley", "1 April, 1978", True, "Gryffindor", "Unknown", "Unknown", "Pure-blood"],
+    ]
+```
+
+
+
+We then define our simple table using `rx.data_editor()`, where we pass in the columns and data as variables.
```
masenf, line 93:
```suggestion
We then define a basic table by passing the previously defined state vars as props `columns` and `data` to the `rx.data_editor()` component, 
```

docs/datatable-tutorial/simple_table.md
```diff
@@ -0,0 +1,104 @@
...
+
+```python demo
+rx.data_editor(
+    columns=DataTableState.cols,
+    data=DataTableState.data,
+)
+```
+
+Now we have our datatable, but there is no way to interact with it. On the next page we will explore how to add interactivity to our datatable.
```
masenf, line 102:
```suggestion
This is enough to display the data, but there is no way to interact with it. On the next page we will explore how to add interactivity to our datatable.
```

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+For our first interactivity example we will use the `on_cell_clicked` event trigger, this is called when the user clicks on a cell of the data editor. The event trigger receives the coordinates of the cell.
+
+```python
+class DataTableState(rx.State):
+    clicked_cell: str = "Cell clicked: "
+
+    ...
+
+
+    def get_clicked_data(self, pos) -> str:
```
masenf, line 28:
```suggestion
    def get_clicked_data(self, pos: tuple[int, int]) -> str:
```

Best to add an annotation here for readability, otherwise i have no idea what this `pos` object looks like

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+
+## Adding Interactivity to our DataTable
+
+
+Now we will add interactivity to our datatable. We do this using event handlers and event triggers.
+
+
+For our first interactivity example we will use the `on_cell_clicked` event trigger, this is called when the user clicks on a cell of the data editor. The event trigger receives the coordinates of the cell.
```
masenf, line 19:
```suggestion
The first example implements a handler for the `on_cell_clicked` event trigger, which is called when the user clicks on a cell of the data editor. The event trigger receives the coordinates of the cell.
```

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+
+    ...
+
+
+    def get_clicked_data(self, pos) -> str:
+        self.clicked_cell = f"Cell clicked: \{pos}"
+        
+```
+
+Here we define a state var called `clicked_cell` that we will use to keep track of the clicked cell. We define an event handler `get_clicked_data` that updates the value of the `clicked_cell` var when it is called. In essence, we have clicked on a cell, called the `on_cell_clicked` event trigger which calls the `get_clicked_data` event handler, which updates the `clicked_cell` var.
```
masenf, line 33:
```suggestion
The state has a var called `clicked_cell` that will store a message about which cell was clicked. We define an event handler `get_clicked_data` that updates the value of the `clicked_cell` var when it is called. In essence, we have clicked on a cell, called the `on_cell_clicked` event trigger which calls the `get_clicked_data` event handler, which updates the `clicked_cell` var.
```

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+```python demo
+rx.data_editor(
+    columns=DataTableState.cols,
+    data=DataTableState.data,
+    on_cell_clicked=DataTableState.get_clicked_data,
+)
+```
+
+
+The event handler `on_cell_context_menu` can be used in the same way as `on_cell_clicked`, except here the event trigger is called when the user right clicks, i.e. when the cell should show a context menu.
```
masenf, line 49: Do we have any example anywhere of displaying a custom context menu? I feel like this is a bit of a tease if we're not going to show the code. And it's a bit more complicated than it sounds here.

At the end of the day, we might leave this line out, or save it for later in the tutorial/API reference.

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Editing cells
+
+Another important type of interactivity we will showcase is how to edit cells. Here we use the `on_cell_edited` event trigger.
```
masenf, line 53:
```suggestion
Another important type of interactivity we will showcase is how to edit cells. Here we use the `on_cell_edited` event trigger to update the data based on what the user entered.
```

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+    clicked_cell: str = "Cell clicked: "
+    edited_cell: str = "Cell edited: "
+
+    ...
+
+
+    def get_clicked_data(self, pos) -> str:
+        self.clicked_cell = f"Cell clicked: \{pos}"
+
+    def get_edited_data(self, pos, val) -> str:
```
masenf, line 66: Suggest more type annotations here.

What is the type of `val`? Is it based on the column type for the table? Is it always a `str` and I have to cast it myself?

docs/datatable-tutorial/simple_table.md
masenf: are there more recognized keys for a column dict as well? where are those defined?

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+        self.data[row][col] = val["data"]
+        self.edited_cell = f"Cell edited: \{pos}, Cell value: \{val['data']}"
+        
+```
+
+The `on_cell_edited` event trigger is called when the user modifies the content of a cell. It receives the coordinates of the cell and the modified content. We pass these into the `get_edited_data` event handler and use them to update the `data` state var at the appropriate position. We then update the `edited_cell` var value.
+
+
+```python demo
+rx.heading(DataTableState.edited_cell)
```
masenf, line 77: might not want to use `heading` here, as these are bigger than the page subheadings and make skimming more difficult

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+```python
+class DataTableState2(rx.State):
+    """The app state."""
+
+    right_clicked_group_header : str = "Group header right clicked: "
+
+    ...
+
+    def get_group_header_right_click(self, index, val):
+        self.right_clicked_group_header = f"Group header right clicked at index: \{index}, Group header value: \{val['group']}"
```
masenf, line 170: not sure i see the utility in this example. it works, but i don't know why someone would want this

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+    on_group_header_context_menu=DataTableState2.get_group_header_right_click,
+)
+```
+
+In this example we use the `on_group_header_context_menu` event trigger which is called when the user right-clicks on a group header. It returns the `index` and the `data` of the group header. We can also use the `on_group_header_clicked` and `on_group_header_renamed` event triggers which are called when the user left-clicks on a group header and when a user renames a group header respectively.
+
+
+### More Event Triggers
+
+There are several other event triggers that are worth exploring. The `on_item_hovered` event trigger is called whenever the user hovers over an item in the datatable. The `on_delete` event trigger is called when the user deletes a cell from the datatable.
```
masenf, line 193: > The `on_delete` event trigger is called when the user deletes a cell from the datatable. 

TIL that deleting a cell was even possible 😁

docs/datatable-tutorial/add_interactivity.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+In this example we use the `on_group_header_context_menu` event trigger which is called when the user right-clicks on a group header. It returns the `index` and the `data` of the group header. We can also use the `on_group_header_clicked` and `on_group_header_renamed` event triggers which are called when the user left-clicks on a group header and when a user renames a group header respectively.
+
+
+### More Event Triggers
+
+There are several other event triggers that are worth exploring. The `on_item_hovered` event trigger is called whenever the user hovers over an item in the datatable. The `on_delete` event trigger is called when the user deletes a cell from the datatable. 
+
+The final event trigger to check out is `on_column_resize`, this allows us to manually resize the width of our columns by hand. The event trigger returns the `col` we are adjusting and the new `width` we have defined. We then index into `self.cols` defined in our state and change the `width` of that column using this code: `self.cols[col['pos']]['width'] = width`.
```
masenf, line 195: > this allows us to manually resize the width of our columns by hand

This sounds sort of funny to me. 

Maybe something more like, "on_column_resize allows us to respond to the user dragging the handle between columns"

I think more description of the `col` arg is warranted, since it appears to be a dictionary, but not clear what is inside.

docs/datatable-tutorial/add_styling.md
```diff
@@ -0,0 +1,143 @@
...
+
+`freeze_columns` is used to keep a certain number of the left hand columns frozen when scrolling horizontally. `group_header_height` and `header_height` define the height of the group header and the individual headers respectively. `max_column_width` and `min_column_width` define how large or small the columns are allowed to be with the manual column resizing. We can also define the `row_height` to make the rows more nicely spaced.
+
+We can add `row_markers`, which appear on the furthest left side of the table. They can take values of `'none', 'number', 'checkbox', 'both', 'clickable-number'`. We can set `smooth_scroll_x` and `smooth_scroll_y`, which allows us to smoothly scroll along the columns and rows. 
+
+By default there is a `vertical_border` between the columns, we can turn it off by setting this prop to `False`. We can define how many columns a user can select at a time by setting the `column_select` prop. It can take values of `"none", "single", "multi"`.
+
+We can allow `overscroll_x`, which allows users to scroll past the limit of the actual horizontal content. There is an equivalent `overscroll_y`. 
+
```
masenf, line 20: Suggest directly linking to the component docs where all of these props (and more) are displayed in table form.

docs/datatable-tutorial/add_styling.md
```diff
@@ -0,0 +1,143 @@
...
+    "fontFamily": "Inter, Roboto, -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Ubuntu, noto, arial, sans-serif",
+}
+```
+
+```python demo
+rx.data_editor(
+    columns=DataTableState2.cols,
+    data=DataTableState2.data,
+
+    #rows=4,
```
masenf, line 119:
```suggestion
```

docs/datatable-tutorial/live_stream.md
```diff
@@ -0,0 +1,112 @@
...
+    "baseFontStyle": "13px",
+    "fontFamily": "Inter, Roboto, -apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Ubuntu, noto, arial, sans-serif",
+}
+```
+
+## Live Streaming example
+
+Lastly let's add in an API so we can live stream data into our datatable.
+
+Here we use a [Background Task](https://reflex.dev/docs/advanced-guide/background-tasks) to stream the data into the table without blocking UI interactivity. We call an advice API using `httpx` and then append that data to the `self.table_data` state var. We also create a button that allows us to start and pause the streaming of the data by changing the value of the boolean state var `running` using the event handler `toggle_pause`. If the `running` state var is set to `True` we stream the API data, when it is set to `False` we break out of the `while` loop and end the background event.
```
masenf, line 37:
```suggestion
Here we use a [Background Task](docs/advanced-guide/background-tasks) to stream the data into the table without blocking UI interactivity. We call an advice API using `httpx` and then append that data to the `self.table_data` state var. We also create a button that allows us to start and pause the streaming of the data by changing the value of the boolean state var `running` using the event handler `toggle_pause`. If the `running` state var is set to `True` we stream the API data, when it is set to `False` we break out of the `while` loop and end the background event.
```

masenf (changes requested):
The conversational tone is a little bit too much for me throughout the tutorial. I think it's great to have a some, to make the reader feel they are part of something, but reading through, i need less "we" and "our" and more "the component ...", "this event handler ...", "the state defines a var, ...", etc.

This would break up the conversational tone with more precise language, without making the whole thing read like reference material.

------------

Adding type hints on all event handlers would help the code examples stand on their own.

docs/datatable-tutorial/add_interactivity.md
tgberkeley: We give an example of on_group_header_context_menu later on. These are not really custom context menus, it is just an event handler that picks up a right mouse click

docs/datatable-tutorial/add_interactivity.md
tgberkeley: it is just an example that shows how to do a right click event handler. I assume people will figure out their own uses for it if they want to ever create events based on right clicks from the user

docs/datatable-tutorial/simple_table.md
tgberkeley: yes this is something I discussed with nikhil, it is just that more commonly people refer to this component as a data table. Perhaps we could call it the data table (editable).

docs/datatable-tutorial/simple_table.md
tgberkeley: as far as I can tell these are the only 4, but going through the reflex docs and the reflex code these are not actually defined anywhere that I can see so it is hard to say if there are more, we will need to chat to Thomas and ask

docs/datatable-tutorial/add_interactivity.md
tgberkeley: added

docs/datatable-tutorial/add_interactivity.md
tgberkeley: val is a dictionary like this: {'kind': 'text', 'data': '20', 'displayData': 'Hermione Jean Granger', 'readonly': False, 'allowOverlay': True} and we index in and just use the 'data' key for now

docs/datatable-tutorial/add_interactivity.md
tgberkeley: changed

docs/datatable-tutorial/add_interactivity.md
tgberkeley: added both of these

docs/datatable-tutorial/live_stream.md
tgberkeley: this actually doesn't seem to work, so changed it back

docs/datatable-tutorial/add_styling.md
tgberkeley: added

PR #306, opened 2023-11-13
substates docs update

ElijahAhianyo:
Docs update  for https://github.com/reflex-dev/reflex/pull/2146

### DIFF
--- a/docs/state/substates.md
+++ b/docs/state/substates.md
@@ -5,52 +5,86 @@ from pcweb.templates.docpage import docdemo_from
 ```
 
 # Substates
-As your app grows, your state will grow too.  You can split your state into
-multiple substates from your base state to keep things organized.
+A substate is a class that inherits from another state class, providing a means for two state classes to share common state resources, such as variables or event handlers.
 
 ## Creating a Substate
-Your base state should inherit from `rx.State`. Substates can either inherit
-from your base state or other substates.
+A substate can typically inherit from a base state(a state that inherits from `rx.State`) or another substate.
+Let’s take a look at a simple example illustrating how and when to use substates. 
+The code below shows two states, `RectangleState` and `SquareState`, each representing metadata about rectangles and squares, along with the calculation of their respective areas.
 
 ```python exec
-class ParentState(State):
-    checked: bool = True
-    count: int = 0
+class RectangleState(rx.State):
+    name: str = "rectangle"
+    shape_type: str = "quadrilateral"
+    length: int = 10
+    width: int = 5
 
+    @rx.var
+    def area(self) -> int:
+        return self.length * self.width 
 
-class ChildState1(ParentState):
-    value: int = 42
+class SquareState(rx.State):
+    name: str = "square"
+    shape_type: str = "quadrilateral"
+    length: int = 4
 
+    @rx.var
+    def area(self) -> int:
+        return self.length ** 2
 
-class ChildState2(ParentState):
-    color: str = "red"
-
-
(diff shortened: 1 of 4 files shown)


ElijahAhianyo:
CI needs https://github.com/reflex-dev/reflex/pull/2146 in next release to pass

docs/state/substates.md
```diff
@@ -5,63 +5,109 @@ from pcweb.templates.docpage import docdemo_from
...
 
-A common use case may be to create a substate for each page of your app, while
-keeping general vars such as the logged in user in the base state for easy
-access.
+
+In this code, you can see that both states share a common resource (`shape_type`). We can improve code organization and promote reusability by introducing the concept of substates. 
+This involves creating a base state (`QuadrilateralState`), from which other states can inherit or subclass. After refactoring, the code takes on the following structure:
+
+```python
+class QuadrilateralState(rx.State):
```
ElijahAhianyo, line 64: The quadrilateral example is a pretty simple one. If there's a better suggestion/example, would be happy to use that

docs/state/substates.md
```diff
@@ -5,63 +5,109 @@ from pcweb.templates.docpage import docdemo_from
...
-from your base state or other substates.
+A substate can typically inherit from a base state(a state that inherits from `rx.State`) or another substate.
+Let’s take a look at a simple example illustrating how and when to use substates. 
+The code below shows two states, `RectangleState` and `SquareState`, each representing metadata about rectangles and squares, along with the calculation of their respective areas.
 
 ```python exec
-class ParentState(State):
-    checked: bool = True
-    count: int = 0
+class RectangleState(rx.State):
```
picklelo, line 16: I think a better example to use here would be something like `HomePageState` and `LoginPageState`. The way it is currently is similar to Python classes/inheritance but the purpose of substates is slightly different, it's more to provide separation of states rather than really doing anything with inheritance.

Basically the way it's structured now evokes the idea that substates need to be similar in terms of what vars they have, which may cause confusion.

docs/state/substates.md
```diff
@@ -5,63 +5,109 @@ from pcweb.templates.docpage import docdemo_from
...
+    length: int = 4
+
+    @rx.var
+    def area(self) -> int:
+        return self.length ** 2
+
+```
+
+It's important to note that certain variables, such as `length` and `name`, were not moved into the `QuadrilateralState` as they are not shared among substates. 
+This approach streamlines the code, ensuring that only common resources are centralized in the base state while preserving the distinct characteristics of individual substates.
```
picklelo, line 87: I think maybe we can have two separate sections. The first section shows just subclassing from `rx.State` for different pages. Then a second example can show subclassing another substate, which will then have this feature of they can share certain variables.

docs/state/substates.md
```diff
@@ -5,63 +5,109 @@ from pcweb.templates.docpage import docdemo_from
...
 
 ## Accessing Parent State Properties
 
 You can access the parent state properties from a child substate, however you
 cannot access the child properties from the parent state.
 
 ```python exec
-def my_heading():
-    return rx.heading(ChildState3.count, color="green")
+def render_substate2():
```
picklelo, line 95: Currently this prints out `render_substate2` in the code, but we should have it say `def index()` instead

picklelo (changes requested):
I'm getting some errors when trying to run this code - does `reflex run` work for you with the latest `main` branch on reflex?

Alek99 (approved):

picklelo (approved):

PR #309, opened 2023-11-14
Memoize markdown components

picklelo:
Speed up compilation - a lot of time was spent compiling out the same markdown components repeatedly.


masenf (approved):
Nice. Takes the compile down from 20s to 11s for me! 😎 And the code is simple.

PR #313, opened 2023-11-15
update the navbar and the docs footer to include a feedback form

tgberkeley:
We still have not linked this up with an external database yet, need to add in the DB URL to check that the data we are sending is being sent to our database

### DIFF
--- a/pcweb/components/navbar.py
+++ b/pcweb/components/navbar.py
@@ -164,7 +164,7 @@ class NavbarState(State):
 
     feedback: str = ""
 
-    score: int = 0
+    page_score: int = 0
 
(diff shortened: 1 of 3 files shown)


Alek99:
I made some more updates for the feedback part on the individual page. I was thinking we could try this https://react-rating-buo7.onrender.com/#custom_box_colors, started to implement the basics but we need to figure out why its not showing up.

I personally prefer the stars if we do keep the faces I would want them to be a black outline with no fill not just the default yellow emojis.

Alek99 (changes requested):
left a comment on next steps

PR #314, opened 2023-11-15
Add integration test to `reflex export`

masenf:
Ensure that the changes can be compiled and exported successfully with the version of reflex in `main`. Allow specifying an alternative reflex dependency for manual runs.


picklelo (approved):
awesome

PR #318, opened 2023-11-17
update sidebar structure

tgberkeley:


masenf (changes requested):
I'm not a big fan of the "Other" categorization. Maybe "Advanced" then then just flatten the "Advanced Guide" section.

Switching to API Reference makes the sidebar wider...it should be fixed width IMO

If i click a link on the API Reference side, the tab should stay on API Reference.

tgberkeley:
I think Advanced works as a title, except then it seems like hosting is advanced, which we do not want it to be.

Also not sure about flattening the advanced guide as these are topics that most users will not look at. We'd be better off flattening the state section or something there

Alek99 (approved):

PR #319, opened 2023-11-20
update gallery for community showcase

tgberkeley:


pcweb/pages/docs/gallery.py
```diff
@@ -196,6 +271,8 @@ def create_list_of_tags(dataframe: pd.DataFrame) -> list:
 class SideBarState(State):
     """Side Bar State"""
 
+    community_apps_list : list[dict[str, str]] = community_apps_list
```
masenf, line 274: Does this have to be state? Seems like we could statically compile it.

tgberkeley: it does so that we can use it in a foreach later on

tgberkeley: also this keeps the structure of the gallery consistent between our gallery and the community gallery, as later on we might want to add the functionality to search through the community gallery based on tags as we currently do for our gallery

PR #323, opened 2023-11-22
add moment docs

Lendemor:


docs/library/datadisplay/moment.md
```diff
@@ -0,0 +1,132 @@
...
+    rx.tab_panels(
+        rx.tab_panel(docdemo(add_example, comp=eval(add_example))),
+        rx.tab_panel(docdemo(subtract_example, comp=eval(subtract_example))),
+    ),
+)
+```
+
+### Timezones
+
+You can also set dates to displays in a specific timezone:
```
masenf, line 128:
```suggestion
You can also set dates to display in a specific timezone:
```

pcweb/component_list.py
```diff
@@ -203,17 +204,17 @@
...
 # The final component list
 component_list = {
-    "Typography": typography_list,
-    "Forms": forms_list,
-    "Layout": layout_list,
-    "Navigation": navigation_list,
+    # "Typography": typography_list,
+    # "Forms": forms_list,
+    # "Layout": layout_list,
+    # "Navigation": navigation_list,
```
masenf, line 210: i think this is probably not meant to be here

docs/library/datadisplay/moment.md
```diff
@@ -0,0 +1,132 @@
...
+```python
+class MomentState(State):
+    date_now: datetime = datetime.now()
+```
+
+### Display the date as-is:
+
+```python eval
+docdemo(
+    as_is_example,
```
masenf, line 76: personally i would `python exec` the examples right above where they are used for better locality and editing.

one of my biggest annoyances with the old docs was having to scroll up to the top to see what `example3` meant; with flexdown we can put the example code and the rendering next to each other

PR #330, opened 2023-12-01
Add copy code button

picklelo:


masenf (commented):
Why not use the can_copy prop of the code_block?

picklelo:
> Why not use the can_copy prop of the code_block?

The default copy button style is not customizable, and I override it to match our website styles. We should probably expose the style props to make the builtin one a bit more usable.

masenf (approved):

PR #334, opened 2023-12-08
conditional rendering first pass

tgberkeley:


docs/conditional_rendering.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+```python eval
+docdemo_from(CondSimpleState, component=cond_simple_example)
+```
+
+
+## Negation
+
+You can use the logical operator `~` to negate a condition. In this example we show that by negating the condition `~CondNegativeState.show` within the cond, we then render the `rx.text("Text 1", color="blue")` component when the state var `show` is negative.
```
Alek99, line 59: Instead of negation this section should have a general title for all var operation and link to the var operation docs as reference

docs/conditional_rendering.md
```diff
@@ -0,0 +1,268 @@
...
+        self.b = random.randint(0, 10)
+        self.c = random.randint(0, 10)
+
+
+def cond_nested_example():
+    return rx.vstack(
+        rx.button("Toggle", on_click=CNS.change),
+        rx.text(f"a: {CNS.a}, b: {CNS.b}, c: {CNS.c}"),
+        rx.cond(
+            ((CNS.a > CNS.b) & (CNS.a > CNS.c)) & ((CNS.a != CNS.b) & (CNS.a != CNS.c)),
```
Alek99, line 216: The combination of nesting and 4 conditionals check is making this example to hard to follow, a more simple example of nesting would make it clearer

Alek99 (changes requested):

docs/conditional_rendering.md
tgberkeley: created a new example above it that is much simpler

Alek99: Nice looks good

Alek99 (approved):

PR #335, opened 2023-12-09
first pass at rendering iterables

tgberkeley:


docs/rendering_iterables.md
```diff
@@ -0,0 +1,234 @@
...
+```python eval
+docdemo_from(SimpleDictForeachState, display_color, component=dict_foreach)
+```
+
+
+
+
+## Nested examples
+
+`rx.foreach` can be used with nested state vars. Here we use nested `foreach` components to render the nested state vars. The `rx.foreach(project["technologies"], get_badge)` inside of the `project_item` function, renders the `dict` values which are `lists`. The `rx.box(rx.foreach(NestedState.projects, project_item))` inside of the `projects_example` function renders each of the `dicts` inside of the overall state var `list`.
```
masenf, line 122:
```suggestion
`rx.foreach` can be used with nested state vars. Here we use nested `foreach` components to render the nested state vars. The `rx.foreach(project["technologies"], get_badge)` inside of the `project_item` function, renders the `dict` values which are `list`. The `rx.box(rx.foreach(NestedState.projects, project_item))` inside of the `projects_example` function renders each `dict` inside of the overall state var `projects`.
```

we should avoid code literal unless the value is the actual exact name of some code you would type... `dict` instead of `dicts` and `list` instead of `lists`

docs/rendering_iterables.md
```diff
@@ -0,0 +1,234 @@
...
+def projects_example() -> rx.Component:
+    return rx.box(rx.foreach(NestedState.projects, project_item))
+```
+
+```python eval
+docdemo_from(NestedState, get_badge, project_item, component=projects_example)
+```
+
+
+If you want an example where not all of the values in the dict are the same type then check out the example on [var operations using foreach]("https://reflex.dev/docs/state/var-operations/#get-item-(indexing)").
```
masenf, line 154:
```suggestion
If you want an example where not all of the values in the dict are the same type then check out the example on [var operations using foreach]("/docs/state/var-operations/#get-item-(indexing)").
```

link should be relative

masenf (changes requested):
looks great tom.

i think we also need a note (alertbox) somewhere on the page that lets people know the type signature of the _functions_ does not matter at all to `foreach`. It's the type annotation on the **state var** that determines what operations are available (e.g. when nesting).

docs/rendering_iterables.md
```diff
@@ -0,0 +1,234 @@
+```python exec
+import reflex as rx
+
+from pcweb.base_state import State
+from pcweb.templates.docpage import docdemo_from, doclink
+```
+
+# Rendering Iterables
+
+You will often want to display multiple similar components from a collection of data. The `rx.foreach` component takes an `iterable(list, tuple or dict)` and a `function` that renders each item in the list. This is useful for dynamically rendering a list of items defined in a state.
```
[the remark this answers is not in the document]
tgberkeley: changed

docs/rendering_iterables.md
```diff
@@ -0,0 +1,234 @@
...
+    return rx.box(rx.text(index), bg=color)
+
+
+def enumerate_foreach():
+    return rx.responsive_grid(
+        rx.foreach(
+            ForeachIndexState.color,
+            lambda color, index: colored_box_index(
+                color, index
+            ),
```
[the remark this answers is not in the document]
tgberkeley: changed

docs/rendering_iterables.md
tgberkeley: changed

docs/rendering_iterables.md
tgberkeley: changed it to use : {var_operations.path}

docs/rendering_iterables.md
```diff
@@ -0,0 +1,234 @@
...
+
+```python eval
+docdemo_from(ForeachIndexState, colored_box_index, component=enumerate_foreach)
+```
+
+
+
+## Dictionary
+
+We can iterate through a `dict` data structure using a `foreach`. When the dict is passed through to the function that renders each item, it is presented as a list of key-value pairs `([1, "blue"],[2, "red"], [3, "green"])`.
```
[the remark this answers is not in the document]
tgberkeley: changed

docs/rendering_iterables.md
```diff
@@ -0,0 +1,241 @@
...
+```python exec
+class NestedDictForeachState(rx.State):
+    color_chart: dict[str, list[str]] = {
+        "purple": ["red", "blue"],
+        "orange": ["yellow", "red"],
+        "green": ["blue", "yellow"],
+    }
+
+
+def display_colors(color: list):
```
masenf, line 174: i wonder if this would look better with a fuller annotation, like `list[str, list[str]]`

masenf (approved):
very helpful docs!

docs/rendering_iterables.md
tgberkeley: updated to include this

PR #375, opened 2024-01-17
Typesense indexing to work again

jackie-pc:
We take the list of routes straight from `pages.routes`, which seems to cover everything.  There is an assumption that every index input is either a flexdown, or a component (which I think is true, or we should make it true).

Another important change is skipping variable interpolation (a.k.a. "Transforms") when parsing flexdown - interpolation only really works during render time (when `python exec` blocks actually get to run - otherwise scope won't contain the variable names needed).

Added some general quality of life things (like nicer logs, `--upload` CLI arg).


scripts/search_index.py
```diff
@@ -47,10 +36,14 @@ def index_flexdown(source: str, href: str) -> list[tuple[str, str, str]]:
 
     # Get the blocks in the source code.
     blocks = flexdown.get_blocks(source)
-    flexdown.process_blocks(blocks, env)
 
     # Get all the markdown blocks.
-    markdown_blocks = [block for block in blocks if block.type == MarkdownBlock.type]
+    markdown_blocks = [block for block in blocks if isinstance(block, MarkdownBlock)]
+
+    # Do not do any transforms - it is only possible during render (which we don't do here).
```
picklelo, line 43: Ah I think maybe we need to `render` the blocks, otherwise the template expressions get left in.

Maybe we can do something similar to the code

```python
# The environment used for execing and evaling code.
env: types.Env = source.metadata

# Get the content of the document.
source = source.content

# Render each block.
markdown_blocks = []
for block in self.get_blocks(source, filename):
    try:
        out.append(block.render(env=env))
        if isinstance(block, MarkdownBlock):
            markdown_blocks.append(block)
    except Exception as e:
        print(
            f"Error while rendering {type(block)} on line {block.start_line_number}. "
            f"\n{block.get_content(env)}"
        )
        raise e
 ```

picklelo (approved):

PR #376, opened 2024-01-17
Fix windows path issue

picklelo:


masenf (approved):
the new-style `pathlib` operations are nicer to work with IMO... but copilot and chatgpt almost always recommend `os.path`, so we end up with a lot of these less-nice APIs used throughout the code.

PR #378, opened 2024-01-18
Edit Radix docs for consistent voice and imports

masenf:
Removing conversational tone in favor of passive voice.

Remove radix shorthand margin and padding props.

Adding `rdxt.` prefix on all radix components for easier find/replace later.


tgberkeley:
This PR looks fantastic, just one minor change from me to a small typo in badge component

ElijahAhianyo:
wondering if we need to have the namespace of `rdxt` defined in the demo block (may be a bit too much for every block ) or have it somewhere in the docs, just so users know what it points to since its entirely different from `rx`

masenf:
@ElijahAhianyo yes i agree. i'm doing it this way for now so that all radix theme and radix primitive components have an explicit prefix so that when we decide how we want to document this later, it's easy to find all the places where we're directly using radix components.


If we do go forward with these changes as they are, then we should definitely be including the suggested imports in an obvious way.

picklelo (approved):
Awesome

PR #382, opened 2024-01-19
Radix Form Docs

martinxu9:
# Summary
- Add Radix form docs with examples
- Requires https://github.com/reflex-dev/reflex/pull/2423


docs/library/forms/form.md
```diff
@@ -0,0 +1,527 @@
...
+                  rdxt.button("Submit"),
+                  as_child=True,
+            ),
+            direction="column",
+            gap="2",
+            align="stretch",
+        ),
+        name="email",
+    ),
+    on_submit=lambda _: rx.window_alert("Submitted!"),
```
picklelo, line 51: We should window alert the form submission: `lambda content: rx.window_alert(content.to_string())`

docs/library/forms/form.md
```diff
@@ -0,0 +1,527 @@
...
+)
+```
+
+A `form_root` contains all the parts of a form. The `form_field`, `form_submit`, etc should all be inside a `form_root`. A `form_field` can contain a `form_label`, a `form_control`, and a `form_message`. A `form_label` is a label element. A `form_control` is where the user enters the input or makes selections. By default, `form_control` is a input. You can use your own form components to construct the `form_control` and set the prop `as_child=True`.
+
+```md alert info
+The current version of Radix Forms does not support composing `form_control` with other Radix form primitives such as `checkbox`, `select`, etc.
+```
+
+The `form_message` is a validation message which is automatically wired (functionality and accessibility). When `form_control` determines the input is invalid, the `form_message` is shown. You set the `match` prop to enable [client side validation](#client-side-validation). You can set the `force_match` prop of `form_message` and `server_invalid` prop of `form_field` together to perform [server side validation](#server-side-validation).
```
picklelo, line 80: This whole section feels very dense and hard to read. It may be okay for now, I think @masenf is going through for voice, but we should simplify this

docs/library/forms/form.md
```diff
@@ -0,0 +1,527 @@
...
+## Data Submission
+
+As previously mentioned, the various pieces of data in the form are submitted together as a dictionary. The form control or the input components must have the `name` attribute. This `name` is the key to get the value from the form data dictionary. If you do not need any validation, you can include the form type components such as Checkbox, Radio Groups, TextArea directly under the `form_root`.
+
+```python demo exec
+import reflex as rx
+import reflex.components.radix.themes as rdxt
+import reflex.components.radix.primitives as rdxp
+
+class RadixFormSubmissionState(rx.State):
```
picklelo, line 95: This whole code can be simplified. We only need one `dict` var for the form data, and can render that.

picklelo (commented):
Docs are very thorough, we just need to make a few sections more digestible. I think most people will use the high level form component so we can make those docs more simple.

docs/library/forms/form.md
martinxu9: is it the excessive use of `code`? I rewrote some of the content by using the proper nouns instead of the code. Also did a first pass using the documentation voice tips from Masen. I feel some of the stuff probably requires a short discussion: the tips are somewhat still up to interpretation. We can do another pass when we scrub the docs

docs/library/forms/form.md
martinxu9: rewrote using the dict and foreach now.

picklelo (approved):

PR #391, opened 2024-01-24
Upgrade to latest flexdown to fix redis issues

picklelo:


pcweb/flexdown.py
```diff
@@ -167,15 +167,13 @@ def render(self, env) -> rx.Component:
         args = lines[0].removeprefix(self.starting_indicator).split()
 
         if "exec" in args:
-            exec(code, env, env)
+            env["__xd"].exec(code, env, self.filename)
```
masenf, line 170: where does `__xd` actually get set? i couldn't find it in the flexdown code

picklelo: Pushed the latest code to Github

Alek99 (approved):
Working well for me

PR #396, opened 2024-01-24
fix gentenv...

jackie-pc:
Have we tried typechecking reflex-web before? :)

### DIFF
--- a/rxconfig.py
+++ b/rxconfig.py
@@ -15,5 +15,4 @@ config = rx.Config(
             "@tailwindcss/typography",
         ],
     },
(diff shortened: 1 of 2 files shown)


rxconfig.py
```diff
@@ -15,5 +15,4 @@
             "@tailwindcss/typography",
         ],
     },
-    redis_url="localhost:6379",
```
jackie-pc, line 18: Assuming this was left in there by accident.

picklelo: good catch

picklelo (approved):

PR #402, opened 2024-01-25
update all form components except textfield to high level low level s…

tgberkeley:
…tructure


docs/library/overlay/dialog.md
```diff
@@ -1,112 +1,13 @@
...
 
-```python exec
-import reflex as rx
-from reflex.components.radix.themes.components import *
-from reflex.components.radix.themes.layout import *
-from reflex.components.radix.themes.typography import *
-```
 
-# Dialog
+# Check Out Low Level Docs! High Level API Coming Soon.
```
masenf, line 12: i'm not really a fan of requiring two clicks to get useful docs

image

If there is no high-level API, can we just not have that tab?

tgberkeley: We can for sure if there is no low level API, but it just looks a little less polished perhaps if there is only one tab for some and two for others. We can discuss and see what we want to do with this though.

tgberkeley: have made a change so that for all docs that do not have high level docs yet will now automatically go to the low level tab

picklelo (approved):

PR #412, opened 2024-01-30
Tom/radix docs updates

tgberkeley:


masenf (commented):
Content looks good to me. I'll have to tweak the names to merge it into my latest PR, but that's fine, i'll deal with that after this lands.

Not sure about the CI failure though, is that just because your change to the event handler rename thing hasn't landed in reflex itself yet?

tgberkeley:
Yes that is the reason I believe, I have to wait for that PR on Reflex to come in first

picklelo (approved):

PR #417, opened 2024-02-01
Handle radix components at top level `rx.` namespace

masenf:
Explicit qualifier for all radix components.

reflex-web is runnable with <del>`reflex-0.4.0`</del> `masenf/namespace-tweaks` branch (https://github.com/reflex-dev/reflex/pull/2523)


picklelo:
Shouldn't we use `rx.` instead of `rdx`?

masenf:
> Shouldn't we use `rx.` instead of `rdx`?

Once your PR lands to move stuff into the top-level, I will update this PR to use those top-level names.

masenf:
The sidebar styling here is absolute bonkers, and the headers look terrible now. A lot of stuff is broken about the website, but i don't want to try to fix those things and conflict with Alek's PR of redesigning the doc pages, so I'll leave it how it is for now.

Any component which used chakra-specific props (like `size="md"` for example), will remain using the chakra component via `rx.chakra.*`. This should make it easy for us to find/replace/fix any outstanding chakra components left in the tree, while keeping the whole site building.

The next step here is to integrate Alek's PR when it's ready (and hope it's using radix components already 😅)

masenf:

PR #431, opened 2024-02-07
document REFLEX_DIR env variable

tgberkeley:


picklelo (approved):
In the future we can probably make a separate reference page for all the env vars we accept

PR #446, opened 2024-02-10
Replace windows style path separator where needed

masenf:
A recent change moved this previous replacement into a function-local variable so it didn't persist outside of that function call.

Therefore, make the replacement again, where needed.

Unbreak windows CI

Testing against live CI in https://github.com/reflex-dev/reflex/pull/2572

### DIFF
--- a/pcweb/pages/docs/__init__.py
+++ b/pcweb/pages/docs/__init__.py
@@ -87,7 +87,10 @@ def get_components_from_metadata(current_doc):
     return components
 
 
-flexdown_docs = flexdown.utils.get_flexdown_files("docs/")
+flexdown_docs = [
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/__init__.py
```diff
@@ -163,7 +163,7 @@ def get_component(doc: str, title: str):
 
 
 for doc in sorted(flexdown_docs):
-    path = doc.split("/")[1:-1]
+    path = doc.replace("\\", "/").split("/")[1:-1]
```
picklelo, line 166: Need to change the route variable below too

picklelo (approved):

PR #452, opened 2024-02-12
form high level docs

martinxu9:
## Summary
- Move original form docs to low level.
- Use the chakra examples as high level form docs.
- Needs https://github.com/reflex-dev/reflex/pull/2594


docs/library/forms/form.md
```diff
@@ -1,497 +1,110 @@
...
-
-The Form Submit is by default a button that submits the form. To use another button component as a Form Submit, include that button as a child inside `form.submit` and set the prop `as_child=True`.
+# Form
 
-The `on_submit` prop of the Form Root accepts an event handler. It is called with the submitted form data dictionary. To clear the form after submission, set the `reset_on_submit=True` prop.
+Forms are used to collect user input. The `rx.form` component is used to group inputs and submit them together.
 
-## Data Submission
+The form component's children can be form controls such as `rx.input`, `rx.checkbox`, or `rx.switch`. The controls should have an `name` attribute that is used to identify the control in the form data. The `on_submit` event trigger submits the form data as a dictionary to the `handle_submit` event handler.
```
tgberkeley, line 14: maybe mention that it can also be others as well eg slider etc

tgberkeley, line 14: The controls should have a `name` attribute

tgberkeley:
All looks good, the only other small thing is that in the API reference it is written as FormRoot in the high level api section, not sure if there is a way to change this to say Form instead?

tgberkeley:
> All looks good, the only other small thing is that in the API reference it is written as FormRoot in the high level api section, not sure if there is a way to change this to say Form instead?

Maybe we can create a new class that is a subclass of form root that we label as form to get the wanted behaviour

martinxu9:
> All looks good, the only other small thing is that in the API reference it is written as FormRoot in the high level api section, not sure if there is a way to change this to say Form instead?

How did we solve for the other components that's alias to the root? I see `hover_card`, `dialog` and `tabs` are all low level only. The other thing we can define is `Form = FormRoot` but I doubt that gives us the props (I think you mentioned that wouldn't work). 
On another note, not sure it's worth this hassle, `form` being `form.root` really has no props to show user anything to change on its visual apparence.

martinxu9:
> > All looks good, the only other small thing is that in the API reference it is written as FormRoot in the high level api section, not sure if there is a way to change this to say Form instead?
> 
> How did we solve for the other components that's alias to the root? I see `hover_card`, `dialog` and `tabs` are all low level only. The other thing we can define is `Form = FormRoot` but I doubt that gives us the props (I think you mentioned that wouldn't work). On another note, not sure it's worth this hassle, `form` being `form.root` really has no props to show user anything to change on its visual apparence.

Never mind. This is resolved in standup.

tgberkeley:
all looks good to me, ready to merge

tgberkeley (approved):

PR #461, opened 2024-02-13
Update chat app example to use rx.form and (radix) rx.input

invrainbow:


masenf (commented):
I think this is an improvement. We should also update the chatapp repo too

https://github.com/reflex-dev/reflex-chat

PR #468, opened 2024-02-13
update indexing script

martinxu9:
## Summary
- Update the component type check in code to the latest component used
- Exclude chakra based components from index.
## Tests
- Run locally without errors. The generated json has no chakra docs.

### DIFF
--- a/scripts/search_index.py
+++ b/scripts/search_index.py
@@ -1,4 +1,5 @@
 """Script to index docs to Typesense."""
+
 import argparse
 import json
 import os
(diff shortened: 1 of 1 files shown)


masenf (approved):

scripts/search_index.py
```diff
@@ -203,6 +205,9 @@ def index_routes():
     """Index the routes."""
     out = {}
     for route in routes:
+        # We do not want to index the docs for Chakra based components.
+        if route.path.startswith("/docs/library/chakra/"):
```
picklelo, line 209: Maybe we should still index them, but add a special tag to them in the document. Then just deprioritize that tag when we're actually doing the query

PR #477, opened 2024-02-15
Fix feedback state and add style updates

Alek99:


Alek99:
i seem to be getting hydration errors on 

http://localhost:3000/docs/state/overview/

Alek99:
Actually now the hydration error are coming up on more pages

PR #488, opened 2024-02-18
restore icon.md

Lendemor:


masenf (changes requested):
Can we update this to just use the `rx.icon` and `rx.flex` in the examples instead of `rdxt`

masenf (approved):

PR #508, opened 2024-03-01
Add "Accessing Arbitrary States" section and example

masenf:


docs/state/overview.md
```diff
@@ -140,12 +140,22 @@ The heading component links its `on_click` event to the
 All interactions between the frontend and backend are handled through events. 
 ```
 
+```md alert info
+# State vs. Instance?
```
picklelo, line 144: nice callout

picklelo (approved):

PR #515, opened 2024-03-05
remove margin around command/output box

martinxu9:


pcweb/templates/docpage/blocks/code.py
```diff
@@ -117,7 +117,6 @@ def doccmdoutput(
         border_radius=styles.DOC_BORDER_RADIUS,
         border=f"2px solid {rx.color('mauve', 3)}",
         position="relative",
-        margin="1em",
```
picklelo, line 120: Do we want to remove, or should this be `margin_y` ?

martinxu9: We'll, remove looks fine to me. We could leave the y margin too

picklelo (approved):

PR #517, opened 2024-03-05
fix links to use hyphen

Lendemor:


pcweb/pages/docs/__init__.py
```diff
@@ -155,14 +155,16 @@ def get_component(doc: str, title: str):
 
 doc_routes = [gallery, library, resources] + apiref_pages
 
-
+t = True
```
picklelo, line 158: leftover?

picklelo (changes requested):
Actually I think the move here is to modify the `.md` file names to always use `-` - so there's consistency between the names and the url. Then maybe we just add an assert here that it doesn't contain an underscore

picklelo (commented):
Looks good, I think we just have to modify how we do the dynamic name spacing now (convert to snake_case before setting it in the namespace)

Lendemor:
> Looks good, I think we just have to modify how we do the dynamic name spacing now (convert to snake_case before setting it in the namespace)

Investigating why but they are already snake_cased in the namespace, yet it still fail.

picklelo (approved):

PR #518, opened 2024-03-06
Website touchups

picklelo:
* add blur to navbar
* make sure copy button highlights correctly
* move telemetry info to FAQ
* Fix some missing headers on right hand sidebar

### DIFF
--- a/pcweb/pages/faq.py
+++ b/pcweb/pages/faq.py
@@ -98,14 +98,12 @@ faq_items = [
         "Q": "What usage data is collected?",
         "A": markdown(
             """
-Anonymous usage data allows us to understand how Reflex is used, what features are most important, and how we can improve the product.
+Anonymous usage data allows us to understand how Reflex is used and how we can improve the product.
 
 The following information is collected:
-* Operating system
-* CPU count
-* Memory
-* Python version
-* Reflex version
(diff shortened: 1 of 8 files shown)


pcweb/pages/faq.py
```diff
@@ -93,6 +94,34 @@
             """
         ),
     },
+    {
+        "Q": "What usage data is collected?",
+        "A": markdown(
+            """
+Anonymous usage data allows us to understand how Reflex is used, what features are most important, and how we can improve the product.
```
Alek99, line 101: Probably need to add the usersid and appid here now

Alek99 (changes requested):

pcweb/templates/docpage/blocks/code.py
```diff
@@ -5,6 +5,20 @@
 from pcweb import styles
 
 
+def copy_icon(code: str) -> rx.Component:
```
Alek99, line 8: With the new refactor the copy_icon seems to be missing for me

Alek99: I was just looking in the learn section, I think here they use the ```python ``` syntax which doesn't use the copy to clipboard, I'll have to fix this in a follow up

PR #542, opened 2024-03-18
add product hunt link

tgberkeley:


pcweb/components/navbar.py
```diff
@@ -513,6 +513,19 @@ def discord_button():
     )
 
 
+def product_hunt_button():
```
masenf, line 516: i think this module is old and should just be deleted

masenf (approved):

masenf (approved):

PR #547, opened 2024-03-21
Update architecture blog

picklelo:


tgberkeley:
For diagram, remove base vars and computed vars and just have vars box

tgberkeley:
maybe look to add numbers (or some way to see the user journey) for the diagram so it is clear where everything starts and finishes

picklelo:
@tgberkeley I'll add these in a follow up don't have time right now

Lendemor (approved):

PR #556, opened 2024-03-24
Fix overlay z-index issue (REF-2346)

markthepixel:
The z-index was not set on the overlay example which made the overlay background fall behind the markdown example code. This fix resolves that issue.

Note: there's still a strange `z_index:50px` attribute in the overlay that both has a wrong case (underscore vs dash) and a value that is in pixels when it shouldn't be


picklelo (approved):
thanks for fixing!

PR #562, opened 2024-03-26
fixed blocking

HongyuHansonYao:
Increase the margin on top, exact margin can be adjust later.

### DIFF
--- a/pcweb/pages/docs/gallery.py
+++ b/pcweb/pages/docs/gallery.py
@@ -531,8 +531,6 @@ def gallery() -> rx.Component:
         ),
         max_width="80em",
         margin_x="auto",
-        margin_top="120px",
+        margin_top="115px",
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/gallery.py
```diff
@@ -531,6 +531,8 @@ def gallery() -> rx.Component:
         ),
         max_width="80em",
         margin_x="auto",
-        margin_top="80px",
+        margin_top="120px",
         height="100%",
     )
+
+# fixme
```
Alek99, line 538: Why is the extra comment here? Is there a todo for later

HongyuHansonYao: Just deleted it, I forget to delete it

picklelo (approved):

PR #566, opened 2024-03-27
Hanson copy button update

HongyuHansonYao:
unnamed
unnamed-1
Improved copy button hover and response functionality


masenf (approved):
i wish we could pop a little tooltip up saying "Code Copied" or something, but that seems really hard given how this is currently implemented...

HongyuHansonYao:
> i wish we could pop a little tooltip up saying "Code Copied" or something, but that seems really hard given how this is currently implemented...
Let's added to the list of ticket, for future!

PR #608, opened 2024-04-10
form_and_dashboard_and_some_spacing

HongyuHansonYao:
I added dashboard and form to the demo section, and adjusted the spacing between sections in landing page.

### DIFF
--- a/pcweb/pages/demos_on_landing_page/dashboard/dashboard.py
+++ b/pcweb/pages/demos_on_landing_page/dashboard/dashboard.py
@@ -1,0 +1,185 @@
+import reflex as rx
+
+data = [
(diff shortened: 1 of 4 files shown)


pcweb/pages/demos_on_landing_page/dashboard/dashboard.py
```diff
@@ -0,0 +1,185 @@
+import reflex as rx
```
picklelo, line 1: We should Cole the top level folder just `landing_demos` or something, keep the names concise (can do in a follow up)

picklelo (approved):
Let's merge it for now and clean up the code next week so people can use it

PR #614, opened 2024-04-15
Hanson mobile view

HongyuHansonYao:
mobile view in progress


Alek99:
Need to remove the red lines, probably used for debugging here

PR #619, opened 2024-04-16
Whitelist for local development only

masenf:
If you actually have changes to the whitelist logic, then you can always add it with `git add -f`, but otherwise, lets avoid accidentally checking in whitelist changes from now on.

### DIFF
--- a/.github/workflows/whitelist.yml
+++ b/.github/workflows/whitelist.yml
@@ -1,0 +1,12 @@
+name: whitelist check
+
+on:
+  pull_request:
+
+jobs:
+  reflex-web:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -13,4 +13,7 @@ __pycache__/
 pcconfig2.py
 pynecone.db
 reflex.db
-venv/
\ No newline at end of file
+venv/
+
+# do not accidentally check in changes to the whitelist used during development
+pcweb/whitelist.py
```
masenf, line 19: bummer, i guess this doesn't actually work 😔

picklelo (approved):

picklelo:
thanks for adding

PR #623, opened 2024-04-16
use on_mount instead of rx.var for fetch components list from CP

martinxu9:
use on_mount instead of rx.var for fetch components list from CP as rx.var fetches the list for every processed event


masenf (approved):
excellent

PR #627, opened 2024-04-16
Custom component gallery

martinxu9:
## Summary
- Fix the icons that are supposed to be links to be links
- Remove the embed/iframe for now

### DIFF
--- a/pcweb/pages/docs/custom_components.py
+++ b/pcweb/pages/docs/custom_components.py
@@ -82,15 +82,12 @@ def updated_on_pypi_if_present(updated_on_pypi: str) -> rx.Component:
 def demo_url_if_present(demo_url: str) -> rx.Component:
     return rx.cond(
         demo_url,
-        rx.link(info_icon(tag="eye"), href=demo_url),
+        rx.link(info_icon(tag="eye"), href=demo_url, is_external=True),
     )
 
 
-def download_url_if_present(download_url: str) -> rx.Component:
-    return rx.cond(
-        download_url,
-        rx.link(info_icon(tag="external-link"), href=download_url),
-    )
+def download_url(download_url: str) -> rx.Component:
+    return rx.link(info_icon(tag="external-link"), href=download_url, is_external=True)
 
 
 def source_if_present(source: str) -> rx.Component:
@@ -99,6 +96,7 @@ def source_if_present(source: str) -> rx.Component:
         rx.link(
             info_icon(tag="github"),
             href=source,
+            is_external=True,
         ),
     )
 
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/custom_components.py
```diff
@@ -186,7 +187,7 @@ def add_item(category: dict) -> rx.Component:
             updated_on_pypi_if_present(category["updated_on_pypi"]),
```
picklelo, line 187: How come this doesn't show up for any of them?

pcweb/pages/docs/custom_components.py
```diff
@@ -186,7 +187,7 @@ def add_item(category: dict) -> rx.Component:
...
             source_if_present(category["source"]),
```
picklelo, line 188: The source link doesn't seem to be working

pcweb/pages/docs/custom_components.py
```diff
@@ -186,7 +187,7 @@ def add_item(category: dict) -> rx.Component:
...
             demo_url_if_present(category["demo_url"]),
```
picklelo, line 189: Can we change this to just be an `rx.link(..., is_external=True)` instead of an iframe for now until we have time to do the iframe styling. We can just leave the iframe commented out for now

picklelo (commented):
The other links need to be added too. Also, can we add the install instructions similar to how we do it on the app gallery? Have a `pip install reflex-X` in a code_block that they can copy

picklelo:
Let me check my code again, I'm seeing the iframe still for some reason

pcweb/pages/docs/custom_components.py
```diff
@@ -81,14 +82,14 @@ def updated_on_pypi_if_present(updated_on_pypi: str) -> rx.Component:
 def demo_url_if_present(demo_url: str) -> rx.Component:
     return rx.cond(
         demo_url,
-        info_icon(tag="external-link"),
+        rx.link(info_icon(tag="eye"), href=demo_url),
```
picklelo, line 85: Let's add `is_external=True` for all the links

pcweb/pages/docs/custom_components.py
martinxu9: because the CP doesn't expose them. I'm releasing a new change from CP in a bit.

pcweb/pages/docs/custom_components.py
```diff
@@ -120,6 +119,15 @@ def pypi_summary(summary: str) -> rx.Component:
     )
 
 
+def pip_install_command_copy_button(package_name: str) -> rx.Component:
+    return rx.link(
+        info_icon(
+            tag="clipboard-copy",
+            on_click=rx.set_clipboard(f"pip install {package_name}"),
```
martinxu9, line 126: is this not a valid var operation? the page doesn't render

martinxu9: @picklelo

picklelo: I think it's a bug, I changed this to be a code block instead

picklelo (approved):

PR #630, opened 2024-04-17
Fix demo

Alek99:
fix forms and dashboard demos on landing page


picklelo (commented):
looks great, just have to undo the whitelist

picklelo (approved):
For the form should we remove the sidebar? Or at least the other options?

PR #640, opened 2024-04-18
Fix example codes in query docs

ElijahAhianyo:


masenf (approved):
hmm not sure how we missed that 😨

PR #642, opened 2024-04-18
requirements.txt: remove email-validator

masenf:
email-validator < 2 is not compatible with pydantic v2

it also doesn't seem like we're using it in reflex-web anymore, so removing the requirement such that reflex-web can be deployed when pydantic v2 is installed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-email-validator==1.3.0
 black==23.10.0
 pandas>=1.5.3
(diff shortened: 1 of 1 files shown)


masenf:
okay, we were using it... not sure why my `git grep` didn't find it at first 😔

martinxu9:
> okay, we were using it... not sure why my `git grep` didn't find it at first 😔 

Me neither. Thought it was not imported

PR #645, opened 2024-04-18
[REF-2536] Use CP backend for source of gallery apps

martinxu9:
Requires https://github.com/reflex-dev/reflex/pull/3113

### DIFF
--- a/pcweb/pages/docs/gallery.py
+++ b/pcweb/pages/docs/gallery.py
@@ -1,343 +1,85 @@
-import pandas as pd
+import json
+import httpx
 
 import reflex as rx
-from pcweb import constants, styles
+
+from rxconfig import config
+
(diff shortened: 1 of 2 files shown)


picklelo (commented):
Everything looks good visually. Only thing is that the page takes a long time for the initial load, I guess it's getting all the images. We may need to add some pagination or scrolling for this to scale.

martinxu9:
> Everything looks good visually. Only thing is that the page takes a long time for the initial load, I guess it's getting all the images. We may need to add some pagination or scrolling for this to scale.
I've set up CDN for all the gallery assets. So for this, it should speed up now. We can chat about other improvements in the app code itself on Monday.

picklelo (approved):

HongyuHansonYao:
Look nice

PR #655, opened 2024-04-23
Custom components gallery updates

HongyuHansonYao:


HongyuHansonYao:
Screenshot 2024-04-24 at 14 27 57
Changes:
1. Changed the hover response for the image from a slight light up effect to a subtle size change, which, in my opinion, looks better.
2. Added a copy button for the command to enhance usability.
3. Made the image clickable, redirecting users to the demo for a more intuitive design.
4. Displayed the number of downloads in the upper right corner to provide relevant information to users.
5. Added the time of the last update in the lower-right corner to keep users informed about the latest changes.
6. Rearranged the entire layout of the page using a grid of components to improve the overall structure and presentation.
7. Performed some code refactoring and cleanup to improve maintainability and readability.

Alek99:
Screenshot 2024-04-24 at 6 09 03 PM

On a thinner screen, it only shows one column, also the spacing when they are two columns is seen to much, they should probably have a three column option. 
Screenshot 2024-04-24 at 6 09 48 PM

Other comments:
Order of copy button and code should be switched
last updates and the date should be on the same line 
The download badge should be at top right there is some space at the top and to the right of it, also should just be gray or a more neutral color. 
Screenshot 2024-04-24 at 6 12 49 PM

HongyuHansonYao:
> Left some commments

Got it!

PR #657, opened 2024-04-24
Improve font section in docs

Alek99:


masenf (approved):
this is significantly better

PR #658, opened 2024-04-24
Docs for style and layout props

tgberkeley:


docs/styling/all-style-and-layout-props.md
Alek99, line 1: The name of this file should be shorter as it going in the sidebar, something like Common Props

docs/styling/all-style-and-layout-props.md
```diff
@@ -0,0 +1,225 @@
...
+            rx.table.column_header_cell(
+                "Description",
+                justify="center",
+                
+            ),
+            rx.table.column_header_cell(
+                "Potential Values",
+                justify="center",
+            ),
+            rx.table.column_header_cell(
```
Alek99, line 211: This column should be centered on the x-axis, align_items="center"

docs/styling/all-style-and-layout-props.md
```diff
@@ -0,0 +1,225 @@
+# All Style and Layout Props
+
+```python exec
+import reflex as rx
+
+props = {
+    "align": {
+        "description": "in a flex, it controls the alignment of items on the cross axis and in a grid layout, it controls the alignment of items on the block axis within their grid area (equivalent to align_items)",
```
Alek99, line 8: The first word in each of the descriptions should be capitalized

docs/styling/all-style-and-layout-props.md
Alek99, line 1: On the top of the document we should add an intro and mention that any CSS props can be used this is just a list of the common ones and link to the general list of CSS props

Alek99 (changes requested):

docs/styling/all-style-and-layout-props.md
Alek99: https://developer.mozilla.org/en-US/docs/Web/CSS

tgberkeley: changed

tgberkeley: added

docs/styling/all-style-and-layout-props.md
tgberkeley: changed

docs/styling/common-props.md
```diff
@@ -0,0 +1,241 @@
+# Style and Layout Props
+
+```python exec
+import reflex as rx
+
+props = {
+    "align": {
```
masenf, line 7: `align` is the radix variant of the prop; i wonder if we should recommend the actual css prop `align_items`

docs/styling/common-props.md
```diff
@@ -0,0 +1,241 @@
...
+        justify="center",
+        align="center",
+        
+    )
+
+```
+
+
+Any [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) prop can be used in a component in Reflex. This is a short list of the most commonly used props. To see all CSS props that can be used check out this [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS).
+
```
masenf, line 210: Should add a line to the effect of `hyphens in CSS property names may be replaced by underscores to use as valid python identifiers`

docs/styling/common-props.md
```diff
@@ -0,0 +1,241 @@
...
+        "description": " Sets the flex grow factor, which specifies how much of the flex container's remaining space should be assigned to the flex item's main size",
+        "values": ["1", "2", "3"],
+        "link": "https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow",
+    },
+    "height": {
+        "description": "Sets an element's height",
+        "values": ["150px", "20em", "75%", "auto"],
+        "link": "https://developer.mozilla.org/en-US/docs/Web/CSS/height",
+    },
+    "justify": {
```
masenf, line 85: similarly here, i wonder if we should list `justify_content`

masenf (approved):
Overall, really excellent reference for new comers and i even learned some new stuff about CSS while reviewing 👏

masenf:
remerge from `origin/main` to pass CI

docs/styling/all-style-and-layout-props.md
tgberkeley: updated

docs/styling/common-props.md
tgberkeley: this is a good point. I do mention that align is equivalent to align_items in the text description

docs/styling/common-props.md
tgberkeley: added

docs/styling/common-props.md
tgberkeley: same as above

docs/styling/common-props.md
```diff
@@ -0,0 +1,225 @@
...
+        "description": "Defines how the browser distributes space between and around content items along the main-axis of a flex container, and the inline axis of a grid container (equivalent to justify_content)",
+        "values": ["start", "center", "flex-start", "space-between", "space-around", "space-evenly", "stretch"],
```
Alek99, line 85: For this section, I see they are in alphabetical order. Should they be grouped by similarity of prop for instance justify and align are very far apart but very similar to each other, so someone who doesn't know the props wont know they are related

Alek99: Same with width/height margin/padding grouping them next to each other seems like it would be more helpful for an end user who doesn't know css

Alek99 (changes requested):

docs/styling/common-props.md
tgberkeley: decided to move onto this in a future PR

PR #659, opened 2024-04-24
Add 404 page only once

ElijahAhianyo:
404 page is already added via `@webpage` decorator in `pcweb/pages/page404`, we shouldn't explicitly add it again via `add_custom_404`

This should unblock https://github.com/reflex-dev/reflex/pull/3155

### DIFF
--- a/pcweb/pages/__init__.py
+++ b/pcweb/pages/__init__.py
@@ -8,7 +8,7 @@ from .index import index
 from .page404 import page404
 
 routes = [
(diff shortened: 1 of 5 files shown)


masenf (changes requested):
we actually need some special logic from add_custom_404_page to make the static redirect work for the deployed site.

specifically `wait_for_client_redirect(self._generate_component(component))`

without that, navigating directly to a url on the deployed site will not actually redirect using client side routing.

masenf (approved):

PR #661, opened 2024-04-24
rearrange logic in FeedbackState.handle_submit

HongyuHansonYao:
where score 1 is thumb up score 0 is thumb down


pcweb/templates/docpage/state.py
Alek99, line 1: Fix formatting and should be good

Alek99 (changes requested):

masenf (approved):

PR #663, opened 2024-04-24
Improve Event Docs Section

Alek99:


docs/events/page_load_events.md
```diff
@@ -19,3 +19,20 @@ class State(rx.State):
 def index():
     return rx.text('A Beautiful App')
 ```
+
+Another example would be checking if the user is authenticated when the page loads. If the user is not authenticated, we redirect them to the login page. If they are authenticated, we redirect don't do anything, letting them access the page. This `on_load` event would be placed on every page that requires authentication to access.
```
masenf, line 23:
```suggestion
Another example would be checking if the user is authenticated when the page loads. If the user is not authenticated, we redirect them to the login page. If they are authenticated, we don't do anything, letting them access the page. This `on_load` event would be placed on every page that requires authentication to access.
```

docs/events/yield_events.md
```diff
@@ -3,22 +3,21 @@ import reflex as rx
...
 
-# Yielding Multiple Updates
+# Yielding Updates
 
 A regular event handler will send a `StateUpdate` when it has finished running. This works fine for basic event, but sometimes we need more complex logic. To update the UI multiple times in an event handler, we can `yield` when we want to send an update.
 
 To do so, we can use the Python keyword `yield`. For every yield inside the function, a `StateUpdate` will be sent to the frontend with the changes up to this point in the execution of the event handler.
 
-```python demo exec
+This example below shows how to yield a 100 updates to the UI.
```
masenf, line 12:
```suggestion
This example below shows how to yield 100 updates to the UI.
```

docs/events/yield_events.md
```diff
@@ -61,3 +60,42 @@ def progress_example():
     )
 
 ```
+
+## Yielding Other Events
+
+Events can also yield other events. This is useful when you want to chain events together. To do this, you can yield the event handler function itself.
+
```
masenf, line 67:
```suggestion
```md alert
# Reference other Event Handler via class

When chaining another event handler with `yield`, access it via the state class, not `self`.
```
```

File: docs/events/event_arguments.md [hunk 4]
```diff
@@ -10,7 +10,7 @@ Try typing a color in an input below and clicking away from it to change the col
 
 ```python demo exec
 class ArgState(rx.State):
-    colors: list[str] = ["rgba(222,44,12)", "white", "#007ac2"]
+    colors: list[str] = ["rgba(245,168,152)", "MediumSeaGreen", "#DEADE3"]
```
masenf, line 13: viva la france 🇫🇷

masenf (changes requested):

masenf (approved):
we should bring this

docs/events/yield_events.md
```diff
@@ -61,3 +60,45 @@ def progress_example():
...
+
+def multiple_yield_example():
+    return rx.cond(
+        YieldEventsState.show_progress,
+        rx.chakra.circular_progress(is_indeterminate=True),
+        rx.heading(
+            YieldEventsState.count,
+            on_click=YieldEventsState.increment,
+            _hover={"cursor": "pointer"},
```
picklelo, line 100: Needs a double curly brace to escape - I'm going to update flexdown to not catch these as it comes up too much.

picklelo (approved):
lgtm

PR #665, opened 2024-04-24
Added a link to default event triggers, used to have nothing

HongyuHansonYao:


pcweb/pages/docs/component.py
```diff
@@ -598,7 +598,12 @@ def generate_event_triggers(comp):
     if not custom_events:
         return rx.vstack(
             rx.heading("Event Triggers", font_size="1em"),
-            rx.text("No component specific event triggers"),
+            rx.link(
+                "See the full list of default event triggers",
+                href="https://reflex.dev/docs/api-reference/event-triggers/",
+                underline="hover",
+                is_external=True,
```
masenf, line 605: Is there a good reason to open in a new tab for this?

HongyuHansonYao: Yes, I think people may want to go back to component docs after see what default trigger are there; and they can easily switch between tabs.

Alek99 (commented):
looks good

PR #668, opened 2024-04-25
Landing page second section

LineIndent:
Started to process figma design onto website. The following pull request is for the landing page, second section with the title "Build web apps, faster...".


HongyuHansonYao:
or we can just put "/" in the whitelist it will only render the landing page

LineIndent:
Is the issue only with the whitelist part? I used `index.py` inside the whitelist list just to speed up compile time and review UI changes quicker. I can remove it and set it to it's default, which was an empty list.

LineIndent:
I've added a new 'task app' to the landing page demo section. Also fixed the title colors and other minor UI things.

PR #669, opened 2024-04-25
Adjust docpage layout, looks better now. Fixed right sidebar overflow

HongyuHansonYao:


masenf (approved):
yeah this looks much nicer

PR #671, opened 2024-04-25
Add redirect docs

tgberkeley:


docs/api-reference/special_events.md
```diff
@@ -53,6 +53,25 @@ When this event is triggered, it navigates the user to a different page or locat
 By default, the redirection occurs in the same tab. However, if you set the external parameter to True, the redirection
 will open in a new tab or window, providing a seamless user experience.
 
+
+This event can also be run from an event handler in State. It is necessary to `return` the `rx.direct()`.
```
masenf, line 57:
```suggestion
This event can also be run from an event handler in State. It is necessary to `return` the `rx.redirect()`.
```

docs/api-reference/special_events.md
```diff
@@ -53,6 +53,25 @@ When this event is triggered, it navigates the user to a different page or locat
...
+
+```python demo exec
+class State(rx.State):
```
masenf, line 60:
```suggestion
class RedirectExampleState(rx.State):
```

docs/library/media/image.md
```diff
@@ -49,3 +49,8 @@ def image_pil_example():
         rx.image(src=ImageState.image)
     )
 ```
+
+```md alert info
+# cv2 image
```
masenf, line 54:
```suggestion
# Using a cv2 image?
```

docs/library/typography/link.md
```diff
@@ -21,6 +22,12 @@ You can also provide local links to other pages in your project without writing
 rx.link("Example", href="/docs/library",)
 ```
 
+```md alert info
+# Redirect the user
+It is also possible to redirect the user to a new path within the application, using `rx.direct()`. Check out the docs [here]({api_reference.special_events.path}).
```
masenf, line 27:
```suggestion
It is also possible to redirect the user to a new path within the application, using `rx.redirect()`. Check out the docs [here]({api_reference.special_events.path}).
```

masenf (changes requested):
A couple of suggestions.

And we should probably just link Events > Special Events to the API Reference > Special Events page which is more complete.

docs/api-reference/special_events.md
tgberkeley: changed

docs/api-reference/special_events.md
tgberkeley: changed

docs/library/media/image.md
tgberkeley: changed

docs/library/typography/link.md
tgberkeley: changed

tgberkeley:
@masenf do you mean remove the Events > Special Events page entirely?

masenf:
> do you mean remove the Events > Special Events page entirely?

Yeah, keep the link, but have it point into the API reference instead. Or we need to drastically improve the page, because right now it's very incomplete.

docs/library/media/image.md
```diff
@@ -49,3 +49,8 @@ def image_pil_example():
         rx.image(src=ImageState.image)
     )
 ```
+
+```md alert info
+# Using a cv2 image?
```
Alek99, line 54: The title of this should be something closer to `Image only accepts URLs and Pillow Images`

as we don't explicitly say what it can take in anywhere.

docs/library/typography/link.md
```diff
@@ -21,6 +22,12 @@ You can also provide local links to other pages in your project without writing
 rx.link("Example", href="/docs/library",)
 ```
 
+```md alert info
+# Redirect the user
```
Alek99, line 26: I think this alert belongs at the bottom of the link docs above style, this spot is a little random for it. Also, title should be more descriptive maybe something like  "Redirecting in within  State"

docs/pages/routes.md
```diff
@@ -133,6 +132,42 @@ This is equivalent to calling `app.add_page` with the same arguments.
...
+
+### Redirect
+
+Redirect the user to a new path within the application using `rx.redirect()`.
+
+- `path`: The destination path or URL to which the user should be redirected.
+- `external`: If set to True, the redirection will open in a new tab. Defaults to `False`.
+
+```python demo
+rx.vstack(
```
Alek99, line 161: The main benefit of redirecting vs a regular link if that you can do it from the state-adding logic behind it. We should show off the difference here so the user has a better idea of what redirect can do,

Alek99 (changes requested):

docs/library/media/image.md
tgberkeley: added

docs/library/typography/link.md
tgberkeley: moved

docs/pages/routes.md
tgberkeley: added

docs/pages/routes.md
```diff
@@ -133,6 +132,56 @@ This is equivalent to calling `app.add_page` with the same arguments.
...
+    rx.button("open in tab", on_click=rx.redirect("/docs/api-reference/special_events")),
+    rx.button("open in new tab", on_click=rx.redirect('https://github.com/reflex-dev/reflex/', external=True))
+)
+```
+
+Redirect can also be run from an event handler in State, meaning logic can be added behind it. It is necessary to `return` the `rx.redirect()`.
```
masenf, line 167: I think showing the logic would be most helpful

```python
class Redirect2ExampleState(rx.State):
    redirect_to_org: bool = False

    def change_page(self):
        if self.redirect_to_org:
            return rx.redirect('https://github.com/reflex-dev/', external=True)
        else:
            return rx.redirect('https://github.com/reflex-dev/reflex/', external=True)
```

tgberkeley: fixed

Alek99 (approved):

PR #678, opened 2024-05-01
Improve API reference section

picklelo:
Add a few more components and improve the layout of the API reference


masenf (commented):
code looks better, but i can't run it; getting the same error as the CI

Alek99 (approved):

PR #682, opened 2024-05-03
add moment docs

Lendemor:


masenf (approved):
Awesome lets get these docs in.

I think adding a section about using `interval` with `on_change` to have a client-side periodic timer could be cool.

Lendemor:
(rerun integration-tests when toast PR is merged)

PR #685, opened 2024-05-04
Alek/galhotfix

Alek99:


picklelo (approved):

pcweb/pages/docs/gallery/state.py
```diff
@@ -56,19 +56,37 @@ def fetch_apps_list(self):
...
+        remaining_apps = []
+        for app in all_apps:
+            if app is not None:
+                # Determine if this app belongs to the community gallery
+                # (for example, you might check `not app.get("is_example_app")`)
+                is_community_app = not app.get("is_example_app")
+
+                if is_community_app:  # Apply the checks only for community apps
+                    none_keys = [k for k, v in app.items() if v is None]
+                    if none_keys:
```
picklelo, line 75: Shouldn't this be inverted? Should we skip if there are none keys?

picklelo (commented):
Might be cleaner to have a function `is_valid_app` and do all the checks in there

PR #687, opened 2024-05-06
Clean up Moment docs

masenf:
* use new-style `python demo` and `python demo exec` blocks for examples
* change `:SS` (fractional part) to `:ss` (seconds)
* make the periodic toast toggleable
* allow `date_now` to be reset to the current time


picklelo (approved):
nice

Lendemor (approved):

PR #689, opened 2024-05-06
fixed alertblock not center, swapped install & learn for onboarding s…

HongyuHansonYao:
1. Fixed some alert block not centering, when alert box has only the body no title, the body should be treated as title, so it aligns to the center. https://linear.app/reflex-dev/issue/REF-2725/alerts-not-centering-heading
2. Swapped install & learn for onboarding section. https://linear.app/reflex-dev/issue/REF-2714/switch-introduction-and-installation-order-on-sidebar
3. On docs/library/forms/select.md, set default to one demo for rx.select doc. https://linear.app/reflex-dev/issue/REF-2778/docslibraryformsselect-simple-styling-demo-have-no-default-selection

### DIFF
--- a/docs/assets/referencing_assets.md
+++ b/docs/assets/referencing_assets.md
@@ -31,8 +31,7 @@ rx.image(src="/Reflex.svg", width="5em")
 ```
 
 ```md alert
-# Always prefix the asset path with a forward slash `/` to reference the asset from the root of the project,
-or it may not display correctly on non-root pages.
+# Always prefix the asset path with a forward slash `/` to reference the asset from the root of the project, or it may not display correctly on non-root pages.
 ```
 
 ## Favicon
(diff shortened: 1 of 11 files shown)


docs/assets/referencing_assets.md
```diff
@@ -31,7 +31,7 @@ rx.image(src="/Reflex.svg", width="5em")
 ```
 
 ```md alert
-Always prefix the asset path with a forward slash `/` to reference the asset from the root of the project,
+# Always prefix the asset path with a forward slash `/` to reference the asset from the root of the project,
```
picklelo, line 34: The line below should also be on the same line, or it makes this a collapsible.

picklelo (commented):
I wonder if instead we can modify the alert block to automatically add a `#` in front if it's a single line?

picklelo (approved):
LGTM - let's make a follow up ticket to see if we can automatically insert the `#` when needed (and remove the accordion in that case)

PR #700, opened 2024-05-14
new errors page

tgberkeley:


masenf:
image

is it supposed to look like this? that text needs more contrast

To a broader point though, i don't really love the format of the page here; it would be much faster and easier to update if these errors were in flexdown blocks IMO.

pcweb/pages/errors.py
```diff
@@ -0,0 +1,270 @@
...
+                    rx.tbody(
+                        rx.tr(
+                            rx.td("foo"),
+                        ),
+                    )
+                )
+                ```
+                """
+            ),
+            rx.text(f"See upstream docs: {rx.link('here.', href='https://nextjs.org/docs/messages/react-hydration-error')}"),
```
masenf, line 81: image

something is wrong here

pcweb/pages/errors.py
```diff
@@ -0,0 +1,270 @@
...
+                """
+                ),
+            rx.markdown(
+                    r"""
+                ```python
+                rx.table(
+                    rx.tr(
+                        rx.td("foo"),
+                    )
+                )
```
masenf, line 54: this is from the old chakra table and won't actually work now, we should update the code to use `rx.table.root`, `rx.table.row` and `rx.table.cell`, etc

masenf (changes requested):

tgberkeley:
Change whole page into flexdown to make it easier to edit

pcweb/pages/errors.py
tgberkeley: fixed this

pcweb/pages/errors.py
tgberkeley: changed

errors.md
```diff
@@ -0,0 +1,135 @@
...
+                rx.text.strong("No ❌"),
+                rx.code_block(
+"""rx.text(rx.text("foo"))
+rx.text(rx.box("foo"))
+rx.text(rx.center("foo"))""",
+                    language="python",
+                ),
+                rx.code_block(
+"""rx.table.root(
+rx.table.row(
```
picklelo, line 79: This should be indented (we can fix in a follow up)

picklelo (approved):

PR #701, opened 2024-05-14
Hanson/news letter and nav menu

HongyuHansonYao:


pcweb/components/docpage/navbar/navbar.py
```diff
@@ -67,82 +68,125 @@ def banner():
...
-            border=f"1px solid {rx.color('mauve', 4)}",
-            background=rx.color("mauve", 1),
+            spacing="6",
             max_width="1000px",
             height="200px",
-            padding="0",
-            overflow="hidden",
         ),
+        border=f"1px solid {rx.color('mauve', 4)}",
+        background=rx.color("mauve", 1),
```
picklelo, line 130: Use color mode cond here

pcweb/pages/index/components/news_letter.py
```diff
@@ -0,0 +1,66 @@
...
+                        background="linear-gradient(180deg, #6151F3 0%, #5646ED 100%)",
+                        box_shadow="0px 2px 9px -4px rgba(64, 51, 192, 0.70), 0px 0px 6px 2px rgba(255, 255, 255, 0.12) inset, 0px 0px 0px 1px rgba(255, 255, 255, 0.09) inset",
+                        border_left="2px solid rgba(186, 199, 247, 0.12)",
+                        border_top_left_radius="0px",
+                        border_bottom_left_radius="0px",
+                        size="lg",
+                        _hover={"linear-gradient(180deg, #6151F3 0%, #5646ED 100%);"},
+                    ),
+                    height="100%",
+                    width="25%",
```
picklelo, line 47: This makes the text overflow in mobile view

picklelo (commented):
We need to add some feedback for when they submit (remove / disable the input and say "You're on the waitlist!"

Alek99 (changes requested):
Nice, only is on the doc page needs to be on the webpages navbar as well and also needs to match the border and border radius of the old on

pcweb/components/docpage/navbar/navbar.py
HongyuHansonYao: Was resolved by set background color transparent

pcweb/pages/index/components/news_letter.py
HongyuHansonYao: I made a ticket for the mobile view, for now, I just disable it on mobile, only shows on tablet and desktop

Alek99 (approved):

PR #713, opened 2024-05-20
Updated rx.text() adding pre-formatting 

HongyuHansonYao:
Some one bought this up on help channel, add one more example.

### DIFF
--- a/docs/library/typography/text.md
+++ b/docs/library/typography/text.md
@@ -175,6 +175,17 @@ rx.text(
 )
 ```
 
+## Preformmatting
+By Default, the browser renders multiple white spaces into one. To preserve whitespace, use `white_space = "pre"` CSS prop.
+
+```python demo
+rx.flex(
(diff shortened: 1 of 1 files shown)


docs/library/typography/text.md
```diff
@@ -175,6 +175,17 @@ rx.text(
 )
 ```
 
+## Preformmatting
+By Default, the browser renders multiple white spaces into one. To preserve whitespace, use `white_space = "pre"` CSS prop.
```
picklelo, line 179: * default should be lowercase
* use "the" CSS prop

docs/library/typography/text.md
```diff
@@ -175,6 +175,17 @@ rx.text(
...
+
+```python demo
+rx.flex(
```
picklelo, line 182: Can use hstack rather than flex column

picklelo (changes requested):

PR #722, opened 2024-05-23
Resolve imports

ElijahAhianyo:


docs/library/datadisplay/list.md
```diff
@@ -1,8 +1,8 @@
 ---
 components:
-    - rx.radix.themes.list.item
-    - rx.radix.themes.list.ordered
-    - rx.radix.themes.list.unordered
+    - rx.radix.list.item
```
picklelo, line 3: Can we event just do `rx.list.item` here?

picklelo (approved):
Looks good - wondering if we should remove the `radix` also and just have the `rx.` in here.

docs/library/datadisplay/list.md
ElijahAhianyo: Yes, we could. I left them in the radix namespace so its explicit that it's a radix component. But I could change them to use the `rx.` namespace if it looks problematic

picklelo: We can do in a follow up if needed

PR #726, opened 2024-05-24
Table docs

tgberkeley:
Updated `rx.table` docs with searching, sorting and pagination for data in state and data from a database. It also includes how to download this data.


docs/library/datadisplay/table.md
```diff
@@ -167,15 +171,548 @@ rx.table.root(
...
+    def prev_page(self):
+        self.offset = max(self.offset - self.limit, 0)
+        self.load_entries()
+
+    def next_page(self):
+        if self.offset + self.limit < self.total_items:
+            self.offset += self.limit
+        self.load_entries()
+
+    def _get_total_items(self, session) -> int:
```
masenf, line 545:
```suggestion
    def _get_total_items(self, session):
```

This does not return an int

masenf (approved):

PR #729, opened 2024-05-25
Catch HTTPError from attempting the Loops request

masenf:
I think the intent was to catch and log errors from the loops requests (for example, missing API key), but RequestError does not catch issues like bad HTTP status that come from `response.raise_for_status()`


masenf:
supersede by #731

PR #735, opened 2024-05-29
EventHandler props can render description from comment

masenf:
Do not render any props starting with `on_` in the main props table.

Extract event trigger descriptions from comments in the source code (where they exist).

### DIFF
--- a/pcweb/pages/docs/component.py
+++ b/pcweb/pages/docs/component.py
@@ -503,6 +503,7 @@ def generate_props(src, component, comp):
         *[
             rx.table.row(*prop_docs(prop, prop_dict, component), align="center")
             for prop in src.get_props()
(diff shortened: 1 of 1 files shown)


pcweb/pages/docs/component.py
```diff
@@ -582,7 +583,10 @@ def same_trigger(t1, t2):
     return args1 == args2
 
 
-def generate_event_triggers(comp):
+def generate_event_triggers(comp, src):
+    prop_name_to_description = {
+        prop.name: prop.description for prop in src.get_props()
```
Lendemor, line 588: I'd expect a `if prop.name.startswith("on_")` to mirror the l506 ? 
Am I missing something obvious ?

Lendemor (approved):

PR #742, opened 2024-06-03
toast docs

Lendemor:
run with PR https://github.com/reflex-dev/reflex/pull/3422

### DIFF
--- a/docs/library/overlay/toast.md
+++ b/docs/library/overlay/toast.md
@@ -1,0 +1,69 @@
+---
+components:
+    - rx.toast.provider
+---
+
+```python exec
+import reflex as rx
+```
+
+# Toast
+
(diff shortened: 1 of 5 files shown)


docs/library/overlay/toast.md
```diff
@@ -0,0 +1,69 @@
...
+
+```python
+rx.button("Show Toast", on_click=rx.toast("Hello, World!"))
+```
+
+## Interaction
+
+If you want to interact with a toast, a few props are available to customize the behavior.
+
+By passing a `ToastAction` to the `action` or `cancel` prop, you can trigger an action when the toast is clicked or when it is closed.
```
masenf, line 36: i think we need an example of how this looks

masenf (approved):

PR #743, opened 2024-06-04
add anchors to errors page

tgberkeley:


picklelo (commented):
Links look good. I think we should remove the Error # in the heading, we don't need to number them. It also will make the links cleaner.

tgberkeley:
Removed error numbers

pcweb/components/webpage/navbar/navbar.py
```diff
@@ -37,88 +37,88 @@ def resources_item(text, url, icon):
...
-            border="1px solid rgba(29, 29, 32, 0.08);",
-            background="linear-gradient(180deg, rgba(29, 27, 33, 0.95) 0%, rgba(20, 19, 24, 0.95) 100%);",
-            box_shadow="0px 24px 54px -17px rgba(13, 12, 16, 0.30), 0px 0px 0px 1px rgba(93, 93, 107, 0.29), 0px 0px 64px 5px rgba(53, 51, 60, 0.30) inset;",
-            max_width="1000px",
-            height="230px",
-            padding="0",
-            overflow="hidden",
-        ),
-    )
+# def resources__section(style):
```
picklelo, line 40: Is this intentional?

tgberkeley: yes as this code was not being used anywhere and was duplicate code

Alek99 (approved):

PR #758, opened 2024-06-13
table blog

tgberkeley:


blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+image: 
+---
+
+```python exec
+from pcweb.pages.docs import library, database
+```
+
+The table component is one of the most useful components for visualizing, editing and working with data. In Reflex we have a built in `rx.table` component. This table efficiently displays data and allows embedding of various Reflex components, such as buttons, dropdowns, checkboxes, or forms, directly within table cells.
+
+In this blog we will link the data in our `rx.table` to an external database. The live version of the app we create in this blog can be found here: https://customer-data-app.reflex.run. The first thing that it is necessary to understand are `Tables`.
```
masenf, line 15:
```suggestion
In this blog we will link the data in our `rx.table` to an external database. The live version of the app we create in this blog can be found here: https://customer-data-app.reflex.run. The first thing that it is necessary to understand are Tables. 
```

We shouldn't use code ticks around things that should not be literally typed as code.

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+    phone: str
+    address: str
+    date: str
+    payments: float
+    status: str
+```
+
+To learn more about databases in Reflex check out the documentation [here]({database.overview.path}).
+
+## Loading Data into the Table
```
masenf, line 38: This is a pretty dense section with a lot of code ticks, it might make sense to break up the explanation with smaller, inline code blocks.

For example, describe the `State` and `load_entries`, then show just that function right below the paragraph.

At the end, you can show the whole code together

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+
+
+
+
+
+## Adding a New Customer 
+
+Next let's explore how to add a user to our database and therefore to our table. We have a function `add_customer_button`, which is a UI element that is rendered in the `main_table` function defined in the section above. 
+
+In this `add_customer_button` function there is an `rx.dialog` which opens when the `rx.button` "Add New Customer" is pressed. When the dialog is opened there is an `rx.form.root` with several `form_field` functions called inside. The `form_field` function returns an `rx.form.field`. It takes in several arguments, but most importantly takes in a `name` argument. The `name` prop is needed to submit with its owning form as part of a `name/value` pair. The content typed into the `rx.input` is submitted as the `value` in that `name/value` pair when the `rx.form` is submitted. The same is done for the `rx.radio` component, where the `name` is set and the radio item selected is passed as the `value` when the `rx.form` is submitted.
```
masenf, line 118:
```suggestion
In this `add_customer_button` function there is an `rx.dialog` which opens when the `rx.button` "Add New Customer" is pressed. When the dialog is opened there is an `rx.form.root` with several `form_field` functions called inside. The `form_field` function returns an `rx.form.field`. It takes in several arguments, but most importantly takes in a `name` argument. The `name` prop is needed to submit with its owning form as part of a name/value pair. The content typed into the `rx.input` is submitted as the `value` in that name/value pair when the `rx.form` is submitted. The same is done for the `rx.radio` component, where the `name` is set and the radio item selected is passed as the `value` when the `rx.form` is submitted.
```

same as above, reserve the code ticks for something that you literally see in the code

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+
+
+
+To submit the `rx.form` with all the `name/value` pairs added, the "Submit Customer" button is pressed which is inside of an `rx.dialog.close` and an `rx.form.submit`, so it closes the dialog and it submits the form. When the form is submitted it runs the code `on_submit=State.add_customer_to_db,` running the event handler `add_customer_to_db`. This event handler sets the `State` var `current_user` to equal the form data just submitted in the `rx.form`. This event handler also does a database query to add the current `self.current_user`, which we just set to be the form response, to the database. Before this a check is done to see if the `email` the user has added is already an `email` in the database, and if it is then an `rx.window_alert` is thrown.
```
masenf, line 122:
```suggestion
To submit the `rx.form` with all the `name/value` pairs added, the "Submit Customer" button is pressed which is inside of an `rx.dialog.close` and an `rx.form.submit`, so it closes the dialog and it submits the form. When the form is submitted it runs the code `on_submit=State.add_customer_to_db`, running the event handler `add_customer_to_db`. This event handler sets the `State` var `current_user` to equal the form data just submitted in the `rx.form`. This event handler also does a database query to add the current `self.current_user`, which we just set to be the form response, to the database. Before this a check is done to see if the `email` the user has added is already an `email` in the database, and if it is then an `rx.window_alert` is thrown. 
```

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+    """The app state."""
+
+    users: list[Customer] = []
+    current_user: Customer = Customer()
+
+    ...
+
+    def add_customer_to_db(self, form_data: dict):
+        """Add a customer to the database."""
+        self.current_user = form_data
```
masenf, line 138: while this mostly works, it probably shouldn't be suggested as it violates the typing of the `current_user` field, which should be a `Customer`, not a `dict`.

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+        self.current_user["date"] = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        with rx.session() as session:
+            if session.exec(
+                select(Customer).where(Customer.email == self.current_user["email"])
+            ).first():
+                return rx.window_alert("User with this email already exists")
+            session.add(Customer(**self.current_user))
```
masenf, line 145:
```suggestion
            self.current_user = Customer(**form_data, date=datetime.now())
            session.add(Customer(self.current_user))
```

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+            session.commit()
```
masenf, line 146:
```suggestion
            session.commit()
            session.refresh(self.current_user)
```

Calling refresh after the submit ensures that the local model gets updated with the inserted ID (in more advanced cases, if any fields are lazy loaded, it would also fetch those so the object is less dependent on the active session and can be serialized)

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+        self.load_entries()
+        return rx.window_alert(f"User \{self.current_user["name"]} has been added.")
+
+
+def form_field(
+    label: str, placeholder: str, type: str, name: str, default_value: str = ""
```
masenf, line 152: even on a very wide display, these long lines overflow the codeblock and require horizontal scroll to see, which is not ideal on desktop

blog/2024-06-14-using-table-component.md
```diff
@@ -0,0 +1,486 @@
...
+            ("Pending", status_badge("Pending")),
+            ("Cancelled", status_badge("Cancelled")),
+            status_badge("Pending")
+        )),
+        rx.table.cell(
+            rx.hstack(
+                update_customer_dialog(user),
+                rx.icon_button(
+                    rx.icon("trash-2"),
+                    on_click=lambda: State.delete_customer(getattr(user, "id")),
```
masenf, line 365:
```suggestion
                    on_click=lambda: State.delete_customer(user.id),
```
did this not work for some reason? should be equivalent and a bit nicer to read

masenf (changes requested):
the post is already quite long and very dense, but it's full of useful techniques and good information that's not really elsewhere in our docs.

as mentioned below, i think the flow could be improved by breaking up the large multi-paragraph explanations with code blocks that pertain to each paragraph (maybe collapsed?) so that i don't lose context by the time i get to see the actual code.

blog/2024-06-14-using-table-component.md
tgberkeley: fixed

blog/2024-06-14-using-table-component.md
tgberkeley: have changed it to take this approach

blog/2024-06-14-using-table-component.md
tgberkeley: fixed

blog/2024-06-14-using-table-component.md
tgberkeley: fixed

blog/2024-06-14-using-table-component.md
tgberkeley: removed

blog/2024-06-14-using-table-component.md
tgberkeley: this has already been added a few lines above here:
```
self.current_user["date"] = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```

blog/2024-06-14-using-table-component.md
tgberkeley: added

blog/2024-06-14-using-table-component.md
tgberkeley: fixed

blog/2024-06-14-using-table-component.md
masenf: the issue i have is that `self.current_user` should be a `Customer`, not a `dict`. As the code is written, it's treating `self.current_user` as a dict, which might be confusing to other event handlers that may want to access `self.current_user`

masenf (commented):
changes mostly look good except for one small issue; we can talk about it in the office tomorrow if you like

masenf (approved):

PR #761, opened 2024-06-16
docs: Add note to select blank template after reflex init in setup tutorial

TG199:
Added a note in the setup tutorial to instruct users to select option 0 (blank template) when prompted during the `reflex init` command. This ensures clarity and helps users set up an empty project correctly.


picklelo (approved):
thanks!

PR #767, opened 2024-06-19
[REF-3144] Add Speed Dial recipes

carlosabadia:


docs/recipes/others/speed_dial.md
```diff
@@ -0,0 +1,471 @@
...
+						"plus",
+						style={
+							"transition": "transform 150ms cubic-bezier(0.4, 0, 0.2, 1)",
+						},
+						class_name="dial",
+					),
+					variant="solid",
+					color_scheme="blue",
+					size="3",
+					cursor="pointer",
```
masenf, line 61: the cursor being a pointer and the rotation that makes the plus look like an "x" makes me think that this button should have something like

```
on_click=cls.toggle(~cls.is_open)
```

This should also allow the element to feel more natural on mobile. Currently tapping the `(+)` works, but then tapping the `(x)` does nothing and i have to tap outside of the component to collapse it.

docs/recipes/others/speed_dial.md
```diff
@@ -0,0 +1,471 @@
...
+					cls.is_open,
+					menu(),
+				),
+				position="relative",
+				style={
+					"&:hover .dial": {
+						"transform": "rotate(45deg)",
+						"transition": "transform 150ms cubic-bezier(0.4, 0, 0.2, 1)",
+					},
+				},
```
masenf, line 75: if you take the previous suggestion, then the rotation state of the element should be based on the value of `is_open`, rather than whether the element is hovered.

So, remove this style dict and add the conditional transform directly to the `icon` component:

```python
rx.icon(
    "plus",
    style={
        "transform": rx.cond(cls.is_open, "rotate(45deg)", "rotate(0)"),
        "transition": "transform 150ms cubic-bezier(0.4, 0, 0.2, 1)",
    },
),
```

docs/recipes/others/speed_dial.md
```diff
@@ -0,0 +1,471 @@
...
+			),
+			on_mouse_over=cls.toggle(True),
```
masenf, line 77: `on_mouse_over` fires every time the mouse moves over an element.

to capture hover/unhover state, use `on_mouse_enter` instead, which only fires once when the mouse first enters the bounds of the element. `on_mouse_enter` is the opposite of `on_mouse_leave` and will provide better performance, particularly on high latency connections where the round trip time is long.

masenf (changes requested):
These components look great. I have a few suggestions that I think are applicable to all of the variants and will improve UX

carlosabadia:
Much cleaner now, thanks @masenf

Alek99:
This is sweet!

masenf (approved):

PR #783, opened 2024-06-26
Update graphing docs tom

tgberkeley:


docs/library/graphing/piechart.md
```diff
@@ -63,55 +121,30 @@ data02 = [
...
-docgraphing(pie_chart_simple_example, comp=eval(pie_chart_simple_example),  data =  "data01=" + str(data01))
-```
-
-We can also add two pies on one chart by using two `rx.recharts.pie` components.
-
-```python eval
-docgraphing(pie_chart_complex_example, comp=eval(pie_chart_complex_example),  data =  "data01=" + str(data01) + "&data02=" + str(data02))
-```
 
 # Dynamic Data
```
Alek99, line 149: This need to be an ## not #

docs/library/graphing/piechart.md
```diff
@@ -149,40 +182,33 @@ class PieChartState(rx.State):
...
-            data=PieChartState.resources,
-            data_key="count",
-            name_key="type_",
-            cx="50%",
-            cy="50%",
-            start_angle=180,
-            end_angle=0,
-            fill="#8884d8",
-            label=True,
+def dynamic_pie_example():
```
Alek99, line 185: This example should have labels or different colors because I can't tell which data corresponds to each quadrant

docs/library/graphing/axis.md
Alek99, line 1: General comment but all the charts should have width set to 100% not a hard pixel number

Alek99: Another General comment the names of the section don't need to have the component in then, see area chart  as to a good naming convention of the sections

Alek99: Naming conversion on the examples some are called line_simple from are called legend_1 we should pick a common convention. I like the descriptive line_simple one

docs/library/graphing/scatterchart.md
Alek99, line 1: I can't see the state code un the dynamic data area chart example

docs/library/graphing/funnelchart.md
Alek99, line 1: The funnel charts event triggers area triggered way to much I suggest we use different ones than on_mouse. We actually can just get rid of this whole section [Example with Event Triggers] because the state example has this

Alek99 (changes requested):

docs/library/graphing/piechart.md
tgberkeley: fixed

docs/library/graphing/piechart.md
tgberkeley: doesn't seem to be a way to do this with the props we currently have

docs/library/graphing/scatterchart.md
tgberkeley: fixed

docs/library/graphing/funnelchart.md
tgberkeley: fixed

docs/library/graphing/axis.md
tgberkeley: All 3 of these have been added as follow up tickets in linear for a future PR

tgberkeley: updated to solve all 3 of these tickets

Alek99 (approved):

PR #786, opened 2024-06-27
[wip] Avoid duplicate state names in markdown files

masenf:
These duplicate names are not causing an obvious error, however when combined with reflex-dev/reflex#3214, the duplicate state names are not resolvable due to how the state classes are redefined from different modules.


benedikt-bartscher:
@masenf I think we can close this since https://github.com/reflex-dev/reflex/pull/3214 is merged

PR #790, opened 2024-07-01
add responsive fields docs

adhami3310:
Changed: /docs/styling/responsive/

### DIFF
--- a/docs/styling/responsive.md
+++ b/docs/styling/responsive.md
@@ -6,7 +6,7 @@ import reflex as rx
 
 Reflex apps can be made responsive to look good on mobile, tablet, and desktop.
 
-You can pass a list of values to any style property to specify it's value on different screen sizes.
+You can pass a list of values to any style property to specify its value on different screen sizes.
 
 ```python demo
 rx.text(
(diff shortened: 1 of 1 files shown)


adhami3310:
The features hasn't been merged for this yet.

docs/styling/responsive.md
```diff
@@ -17,14 +17,94 @@ rx.text(
...
+        ),
+        rx.table.row(
+            rx.table.cell("xl"),
+            rx.table.cell("96em"),
+        ),
+    ),
+)
+```
+
+You can cusomize them using the style property.
```
masenf, line 104:
```suggestion
You can customize them using the style property.
```

docs/styling/responsive.md
```diff
@@ -17,14 +17,94 @@ rx.text(
 
 The text will change color based on your screen size. If you are on desktop, try changing the size of your browser window to see the color change.
 
+You can also use `rx.breakpoints`. Each size maps to a corresponding key, the value of each will be applied when the screen size is greater than or equal to the named breakpoint.
```
masenf, line 20:
```suggestion
_New in 0.5.6_

Responsive values can also be specified using `rx.breakpoints`. Each size maps to a corresponding key, the value of which will be applied when the screen size is greater than or equal to the named breakpoint.
```

docs/styling/responsive.md
```diff
@@ -17,14 +17,94 @@ rx.text(
...
+    "Hello World",
+    color=rx.breakpoints(
+        initial="orange",
+        sm="purple",
+        lg="green",
+    ),
+)
+```
+
+You can also specify custom values per component using a dictionary instead of named parameters.
```
masenf, line 33:
```suggestion
Custom breakpoints in CSS units can be mapped to values per component using a dictionary instead of named parameters.
```

Personally I think avoiding the "You can ..." phrasing makes the documentation sound more authoritatize and attributes the capability/behavior to the feature rather than the user. Sometimes the "You can ..." phrasing makes sense, and is used in other pages of the docs, but generally should be avoided.

masenf (changes requested):

docs/styling/responsive.md
adhami3310: That makes sense, will try to stick to that in the future!

masenf (approved):

PR #791, opened 2024-07-01
Initial video block 

tgberkeley:


tgberkeley:
`title` is rendered on line 304 of flexdown.py `rx.markdown(title) if title else rx.markdown("Video Description"),` 
`content` is no longer needed I agree. Ideally want to keep this all styled like the old info blocks

PR #799, opened 2024-07-05
Fix CopyPasta in signup_form.md 

vo55:
In the signup form recipe, change the buttont text from "sign in" to "register" as this seems to be a copypasta mistake


picklelo (approved):
thanks!

PR #804, opened 2024-07-12
Fix sidebar header links going to overview

picklelo:


pcweb/components/docpage/sidebar/sidebar.py
```diff
@@ -413,10 +413,10 @@ def sidebar_comp(
                 3,
                 rx.flex(
                     create_sidebar_section(
-                        "Reference", api_reference, api_reference_index, url
+                        "Reference", pages[0].path, api_reference, api_reference_index, url
                     ),
                     create_sidebar_section(
-                        "Tutorials", tutorials, tutorials_index, url
+                        "Tutorials", pages[0].path, tutorials, tutorials_index, url
```
masenf, line 419: both of these link to the autogenerated API docs for `App`, but that doesn't seem quite right.

The Reference link should probably go to the first page in the reference section, which is CLI.

The Tutorials link should probably go to the first page in the datatable tutorial.

Alek99 (approved):

PR #818, opened 2024-07-18
add docs for exception handler

Lendemor:


docs/utility_methods/exception_handlers.md
```diff
@@ -0,0 +1,40 @@
+# Exception handlers
+
+_Added in v0.5.7_
+
+Exceptions handlers are function that can be assigned to your app to handle exceptions that occur during the application runtime.
```
picklelo, line 5: function -> functions

docs/utility_methods/exception_handlers.md
```diff
@@ -0,0 +1,40 @@
...
+They are useful for customizing the response when an error occurs, logging errors, and performing cleanup tasks.
+
+## Types
+
+Reflex support two type of exception handler `frontend_exception_handler` and `backend_exception_handler`.
```
picklelo, line 10: handler -> handlers

docs/utility_methods/exception_handlers.md
```diff
@@ -0,0 +1,40 @@
...
+
+They are both used to handle exceptions that occur in the `frontend` and `backend` respectively.
```
picklelo, line 12: remove "both". Maybe explain a bit what a frontend error is compared to a backend. A frontend is a Javascript error, while a backend is an error in the event handler

docs/utility_methods/exception_handlers.md
```diff
@@ -0,0 +1,40 @@
...
+```md alert warning
+# Only named functions are supported as exception handler.
+```
+
+## Examples
+
+```python
+import reflex as rx
+
+def custom_frontend_handler(exception: Exception) -> None:
```
picklelo, line 29: Let's add some examples of actual handlers here to show how it could work

picklelo (commented):
Looks good, some minor nits.

I'm wondering though (outside of this PR) should exception handlers be tied to event handlers? then for example you could update your state if something happened.

docs/utility_methods/exception_handlers.md
Lendemor: What do you have in mind for actual handlers? Printing the exception ?

picklelo: Yeah I think printing it makes sense, just to show a practical way someone may use it

docs/utility_methods/exception_handlers.md
```diff
@@ -0,0 +1,42 @@
...
+    # My custom logic for frontend errors
+    pass
+
+def custom_backend_handler(exception: Exception) -> rx.event.EventSpec:
+    # My custom logic for backend errors
+    pass
+
+app = rx.App()
```
picklelo, line 39: Try this with the kwargs approach instead

picklelo (approved):

PR #823, opened 2024-07-20
improve toast docs

Lendemor:


docs/library/overlay/toast.md
```diff
@@ -66,4 +66,21 @@ rx.button(
         style={"background-color": "green", "color": "white", "border": "1px solid green"}
     )
 )
-```
\ No newline at end of file
+```
+
+The following props are available for customization:
```
picklelo, line 71: Shouldn't these get auto-documented at the bottom? I think we can remove all this

Lendemor: They're not component, so no.

We'd need to add additional logic to auto-document any class inheriting from `PropsBase`.

picklelo (approved):

PR #824, opened 2024-07-20
add titles for overview sections

tgberkeley:


pcweb/pages/docs/__init__.py
```diff
@@ -122,7 +138,11 @@ def get_component(doc: str, title: str):
     # Get the docpage component.
     doc = doc.replace("\\", "/")
     route = rx.utils.format.to_kebab_case(f"/{doc.replace('.md', '/')}")
-    title2 = to_title_case(title)
+    if doc in manual_titles.keys():
+        title2 = manual_titles[doc]
+        print(title2)
```
Lendemor, line 143: debug print ?

pcweb/pages/docs/__init__.py
```diff
@@ -112,6 +113,21 @@ def exec_blocks(doc, href):
 outblocks = []
 
 
+manual_titles = {
```
Lendemor, line 116: We should put each title in the meta block on top of each md file instead of here.

pcweb/pcweb.py
```diff
@@ -59,6 +59,7 @@
 
 # Add the pages to the app.
 for route in routes:
+    print(route.title)
```
Lendemor, line 62: debug print

Lendemor (changes requested):

pcweb/pages/docs/__init__.py
```diff
@@ -112,6 +113,21 @@ def exec_blocks(doc, href):
...
+    "docs/library/datadisplay/callout-ll.md": "Callout Low",
```
picklelo, line 117: I think in the place where we set the low level, we can just modify the name, then it will apply to all Low Level components.

pcweb/pages/docs/__init__.py
```diff
@@ -112,6 +113,21 @@ def exec_blocks(doc, href):
...
+    "docs/database/overview.md": "Database Overview",
+    "docs/custom-components/overview.md": "Custom Components Overview",
+    "docs/api-routes/overview.md": "API Routes Overview",
+    "docs/client_storage/overview.md": "Client Storage Overview",
+    "docs/state/overview.md": "State Overview",
+    "docs/styling/overview.md": "Styling Overview",
```
picklelo, line 123: Similarly here - I think we have some logic on the prev/next buttons to modify any Overview page with the correct title. Maybe we can reuse the same function here

pcweb/pcweb.py
picklelo, line 62: remove before merging

pcweb/pages/docs/__init__.py
tgberkeley: removed

pcweb/pcweb.py
tgberkeley: removed

pcweb/pages/docs/__init__.py
tgberkeley: changed

pcweb/pcweb.py
tgberkeley: removed

picklelo (approved):

pcweb/pages/docs/__init__.py
picklelo: agreed let's do this in the future

PR #835, opened 2024-08-01
 Add docs for HTML Components

Alek99:
Up for debate if we want to separate this out into other page. More pushing this as a solution if they already know html rather than a guide on how to use html.


Alek99:
Needs this pr to pass test https://github.com/reflex-dev/reflex/pull/3731/files

picklelo (approved):
Docs look good 

I noticed some of the docstrings are incorrect, i.e. all the h1 h2 h3 say "displays an h1 element" but that's on the reflex side.

PR #841, opened 2024-08-07
Fix: Correct 'Edit this page' link in documentation footer

doshikaran:
This PR fixes the issue with the "Edit this page" link at the bottom of the documentation pages, which was previously leading to a 404 error. The link was incorrectly formatted, causing it to point to a non-existent path in the GitHub repository.

**Changes made:**
Updated the docpage_footer function to correctly format the path for the "Edit this page" link.

### DIFF
--- a/pcweb/templates/docpage/docpage.py
+++ b/pcweb/templates/docpage/docpage.py
@@ -97,6 +97,8 @@ def docpage_footer(path: str):
     from pcweb.pages.docs.gallery import gallery
     from pcweb.pages.docs import getting_started, hosting
     from pcweb.pages.changelog import changelog
+
+    
     return rx.flex(
         rx.divider(size="4"),
         rx.flex(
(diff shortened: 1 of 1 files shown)


pcweb/templates/docpage/docpage.py
```diff
@@ -153,7 +155,9 @@ def docpage_footer(path: str):
                             padding="0px 10px",
                             white_space="nowrap",
                         ),
-                        href=f"https://github.com/reflex-dev/reflex-web/tree/main{path}.md",
+                        href=f"https://github.com/reflex-dev/reflex-web/blob/main/docs/getting-started/installation.md",
```
picklelo, line 158: We shouldn't hardcode this path since it changes for every dockage. It should instead reference the path. I think all we need to do is replace the original with `path.strip('/')` to strip of the extra trailing slash

picklelo (changes requested):

PR #842, opened 2024-08-09
Fix:Correct 'Edit this page' link in documentation footer

parthamv:
I have added a normalised_path variable to redirect the trailing slashes request as well to the correct path and making it generic

### DIFF
--- a/pcweb/templates/docpage/docpage.py
+++ b/pcweb/templates/docpage/docpage.py
@@ -98,9 +98,8 @@ def docpage_footer(path: str):
     from pcweb.pages.docs import getting_started, hosting
     from pcweb.pages.changelog import changelog
 
-    path = str(path)
-    if len(path) > 0 and path[-1] == '/':
-        path = path[:-1]
+    
+    path = path[:-1]
     return rx.flex(
         rx.divider(size="4"),
         rx.flex(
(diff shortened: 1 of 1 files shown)


reflex-web
Alek99, line 1: Why is a submodule added

parthamv: I have removed the submodule and have just stripped the last character

parthamv:
I have removed the submodules and I have added truncated and resolved the path var issue

pcweb/templates/docpage/docpage.py
```diff
@@ -97,6 +97,9 @@ def docpage_footer(path: str):
     from pcweb.pages.docs.gallery import gallery
     from pcweb.pages.docs import getting_started, hosting
     from pcweb.pages.changelog import changelog
+
+    
+    path = path[:-1]
```
picklelo, line 102: We should do `path.strip("/")` just in case we pass one that doesn't have the trailing slash

picklelo (commented):
thanks! one small comment

pcweb/templates/docpage/docpage.py
parthamv: path.strip("/") is not working as the path is a baseVar in the state, hence I have done a work around to check if there is a trailing slash using rx.cond and then have removed the last one if there is one. The problem seems to have been resolved hence it will handle both the case when there is no trailing slash and when there is one

parthamv:
Hi I have made the changes I guess if someone can review and resolve this pull request

PR #848, opened 2024-08-14
Update hosting-cli-commands.md

stock90975:
"...to automatically detects your existing app by the same name..." 

should be:

"... to automatically detect your existing app with the same name..."


picklelo (approved):
thanks!

PR #912, opened 2024-09-16
update event args docs

tgberkeley:


docs/pages/dynamic_routing.md
```diff
@@ -1,83 +1,64 @@
...
+```python
 
-dynamic_routes = (
-"""
 class State(rx.State):
     @rx.var
     def post_id(self) -> str:
+        # Retrieves the value of the 'pid' parameter from the current route
+        # If 'pid' is not found, it defaults to 'no pid'
         return self.router.page.params.get('pid', 'no pid')
```
masenf, line 26: is there a reason why we get the arg this way instead of via `self.pid`?

all dynamic route args should be accessible via constructs like `rx.State.arg_name` and `self.arg_name`

PR #929, opened 2024-09-20
v0.6.0 blog post

Alek99:


blog/2024-09-21-reflex-v0.6.0.md
```diff
@@ -0,0 +1,128 @@
...
+        width="100%",
+    ),
+    width="100%",
+)
+```
+
+
+### Responsive Support For All Props
+
+Reflex now comes with configurable responsive breakpoints for all props. If you change the window size the component will update to match the new breakpoint, see the example below.
```
masenf, line 86: I think we still only support style props for rx.responsive

blog/2024-09-21-reflex-v0.6.0.md
```diff
@@ -0,0 +1,128 @@
...
+    {"name": "F", "uv": 40, "pv": 170, "amt": 250},
+    {"name": "G", "uv": 50, "pv": 190, "amt": 260},
+]
+```
+
+Reflex v0.6.0 is here! In this blog post, we’ll go over all the major improvements and changes since v0.5.0.
+
+### Decentralized Var Operations
+
+In this release, we reworked vars and made var operations to be more extensible. These changes make vars more stable and easier to work with along with adding more type safety.
```
masenf, line 32: If we're going to lead with this, we might need a sentence explaining what a var is in this context

blog/2024-09-21-reflex-v0.6.0.md
```diff
@@ -0,0 +1,128 @@
...
+
+
+To an end reflex users, there is no visible change and it should be backward compatible with previous versions.
```
masenf, line 35: We should link to the docs where we explain that Var.create makes a python value into a js literal. And Var() creates a value that renders directly in the generated JS. Noting that is_local and is_string are deprecated now.

docs/api-reference/var_system.md
```diff
@@ -0,0 +1,82 @@
...
+
+State fields are converted to `Var` by default. Additionally, you can create a `Var` from Python values using `rx.Var.create()`:
+
+```py
+rx.Var.create(4) # NumberVar
+rx.Var.create("hello") # StringVar
+rx.Var.create([1, 2, 3]) # ArrayVar
+```
+
+If you want to explicitly create a `Var` from a raw Javascript string, you can call the init function on `rx.Var`:
```
masenf, line 46:
```suggestion
If you want to explicitly create a `Var` from a raw Javascript string, you can instantiate `rx.Var` directly:
```

docs/api-reference/var_system.md
```diff
@@ -0,0 +1,82 @@
...
+rx.Var("2", _var_type=int).guess_type() # NumberVar
+```
+
+In the example above, `.guess_type()` will attempt to downcast from a generic `Var` type into `NumberVar`.
+For this example, calling the function `.to(int)` can also be used in place of `.guess_type()`.
+
+## Operations
+
+The `Var` system also supports some other basic operations.
+For example, `NumberVar` supports basic arithmetic operations like `+` and `-` like in Python.
```
masenf, line 58:
```suggestion
For example, `NumberVar` supports basic arithmetic operations like `+` and `-`, as in Python.
```

masenf (approved):
rebased and cleaned up commit history. i'm good with merging this now

PR #947, opened 2024-09-24
piechart colors

Lendemor:
Closes https://github.com/reflex-dev/reflex/issues/2080


docs/library/graphing/charts/piechart.md
```diff
@@ -26,23 +25,28 @@ data01 = [
   },
   {
     "name": "Group B",
-    "value": 300
+    "value": 300,
+    "fill":"#D41009FF"
```
Alek99, line 29: Can we use rx.color here instead

Alek99: for all the colors

Alek99 (changes requested):

docs/library/graphing/charts/piechart.md
Lendemor: I wanted to use different colors to show we support all of them.
But I can make that change.

Alek99 (approved):

PR #1028, opened 2024-10-14
Alek/formdocs

Alek99:


docs/library/forms/radio_group.md
tgberkeley, line 1: We should also show examples of how to us the `value` and `default_value` props

picklelo (commented):
I think we should standardize on hyphen-case for the markdown files. Some are snake_case now and it's inconsistent

picklelo (approved):

PR #1056, opened 2024-10-28
[ENG-4001] Doc bash fixes for UI section

picklelo:


docs/components/conditional_rendering.md
tgberkeley, line 1: I am wondering where we put all the examples you deleted in this page?

docs/components/props.md
```diff
@@ -1,30 +1,30 @@
...
+Check the docs for the component you are using to see what props are available and how they affect the component (see the `rx.image` [reference]({docs.library.media.image.path}#api-reference) page for example).
 
-```md alert success
-# Reflex has a wide selection of [built-in components]({library.path}) to get you started quickly.
-```
 
-## HTML Props
+## Common Props
 
 Components support many standard HTML properties as props. For example: the HTML [id]({"https://www.w3schools.com/html/html_id.asp"}) property is exposed directly as the prop `id`. The HTML [className]({"https://www.w3schools.com/jsref/prop_html_classname.asp"}) property is exposed as the prop `class_name` (note the Pythonic snake_casing!).
```
tgberkeley, line 30: might be nice for this example to actually define class-name-1 and class-name-2 so it is a bit more tangible what is happening here, as people coming from a non-web dev background might be like what is happening in this example?

docs/components/rendering_iterables.md
```diff
@@ -1,46 +1,106 @@
...
-        columns="6",
     )
 
 ```
 
-```md alert warning
-# The type signature of the functions does not matter to the `foreach` component.
+In this first simple example we iterate through a `list` of colors and render a dynamic number of buttons.
+
+The first argument of the `rx.foreach` function is the state var that you want to iterate through. The second argument is a function that takes in an item from the data and returns a component. In this case, the `colored_box` function takes in a color and returns a button with that color.
```
tgberkeley, line 33: for state var here let's link back to the state docs again

tgberkeley, line 33: here when we say "the second argument" line, maybe we mention that the data is an iterable, i.e. like a list or a dict etc and this function takes each item individually from this list and passes it through that function.

tgberkeley, line 33: and for the line where we discuss the color, maybe we can say the colored_box function takes in each color individually from the list that is the State var called color and returns a button with that color.

docs/components/rendering_iterables.md
```diff
@@ -1,46 +1,106 @@
...
+
+## For vs Foreach
+
+```md definition
+# Regualar For Loop
```
tgberkeley, line 38: Regular

docs/components/rendering_iterables.md
```diff
@@ -1,46 +1,106 @@
...
+* Use when iterating over state vars.
+```
+
+The above example could have been written using a regular Python `for` loop, since the data is constant.
+
+```python demo exec
+colors = ["red", "green", "blue"]
+def dynamic_buttons_for():
+    return rx.vstack(
+        [colored_box(color) for color in colors],
```
tgberkeley, line 51: how does this line work without the * in the list comprehension

docs/components/rendering_iterables.md
```diff
@@ -1,46 +1,106 @@
...
+def dynamic_buttons2():
+    return rx.vstack(
+        rx.foreach(IterState2.color, colored_box),
+    )
 
-Rather the type annotation on the `state var` determines what operations are available (e.g. when nesting).
 ```
 
-## Enumeration
+Notice that the type annotation for the `color` parameter in the `colored_box` function is `rx.Var[str]` (rather than just `str`). This is because the `rx.foreach` function passes the item as a `Var` object, which is a wrapper around the actual value. This is what allows us to compile the frontend without knowing the actual value of the state var (which is only known at runtime).
```
tgberkeley, line 101: maybe a link for the runtime line to: docs/getting-started/basics/#compile-time-vs.-runtime-(important)

docs/components/rendering_iterables.md
```diff
@@ -1,46 +1,106 @@
...
         "red",
         "green",
         "blue",
-        "yellow",
-        "orange",
-        "purple",
     ]
 
 
 def colored_box(color: str):
```
tgberkeley, line 20: how come this example has the typing as `str` but later under the render function section we specifically state that "type annotation for the color parameter in the colored_box function is rx.Var[str] (rather than just str). "

docs/components/rendering_iterables.md
tgberkeley, line 1: I think it is confusing that some examples are using a lambda and other a separately defined function. I think in the docs we should only show one way to do it which should be the defining a separate function. Where we mention that you can do lambda functions we should show one easy example (the same color example with a lambda) to show what it looks like and that is it.

docs/components/rendering_iterables.md
```diff
@@ -1,46 +1,106 @@
...
 
-Rather the type annotation on the `state var` determines what operations are available (e.g. when nesting).
 ```
 
-## Enumeration
+Notice that the type annotation for the `color` parameter in the `colored_box` function is `rx.Var[str]` (rather than just `str`). This is because the `rx.foreach` function passes the item as a `Var` object, which is a wrapper around the actual value. This is what allows us to compile the frontend without knowing the actual value of the state var (which is only known at runtime).
+
+## Enumerating Iterables
 
 The function can also take an index as a second argument, meaning that we can enumerate through data as shown in the example below.
```
tgberkeley, line 105: the Render function

docs/components/rendering_iterables.md
```diff
@@ -50,26 +110,25 @@ class IterIndexState(rx.State):
...
+                padding_y="0.5em",
+            ),
         ),
-        columns="6",
     )
 
 ```
 
-## Dictionary
+## Iterating Dictionaries
```
tgberkeley, line 129: might be nice to mention the necessity of typing correctly here so that foreach works

docs/components/rendering_iterables.md
```diff
@@ -125,7 +184,7 @@ def projects_example() -> rx.Component:
     return rx.box(rx.foreach(NestedStateFE.projects, project_item))
 ```
 
-If you want an example where not all of the values in the dict are the same type then check out the example on [var operations using foreach]({vars.var_operations.path}).
+If you want an example where not all of the values in the dict are the same type then check out the example on [var operations using foreach]({docs.vars.var_operations.path}).
```
tgberkeley, line 187: I think this is so so important there should be an example here of using `rx.Base`. It is the biggest question we get again and again with dict with values of different types. This is something we must add in these docs, not just a link.

docs/pages/overview.md
```diff
@@ -147,13 +153,55 @@ app = rx.App()
```
tgberkeley, line 153: for the example above I think we should use add_page if this is the approach we are trying to push (again I think we should have one way to do it and push them in the docs, and then we can show an alternative, but it can get confusing if you are showing two ways to do it everywhere)

docs/pages/overview.md
```diff
@@ -163,6 +211,13 @@ class State(rx.State):
         # ... Your logic here ...
 ```
 
+The `router.page.path` attribute allows you to obtain the path of the current page from the router data,
+for dynamic pages this will contain the slug rather than the actual value used to load the page.
```
tgberkeley, line 215: link to dynamic pages docs

docs/pages/overview.md
```diff
@@ -178,15 +233,3 @@ class State(rx.State):
 ```
 
 In this example, running on `localhost` should display `http://localhost:3000/user/hey/posts/3/`
```
tgberkeley, line 235: I am not sure what example this is referring to?

docs/components/rendering_iterables.md
ElijahAhianyo: We can stick to using a separate function but for some cases, maybe it would make sense to show what using a lambda would look like. Typical example is the section on enumerating iterables where there's an index `arg`. 
With the lamda, the index `arg` is explicitly stated whereas using with the case of defining a function, an extra arg/param would represent the index. Code below

Using Lamda:

```python
class IterIndexState(rx.State):
    color: list[str] = [
        "red",
        "green",
        "blue",
    ]


def enumerate_foreach():
    return rx.vstack(
        rx.foreach(
            IterIndexState.color,
            lambda color, index: rx.box(
                rx.button(f"{index + 1}. {color}"),
                padding_y="0.5em",
            ),
        ),
    )

```

Using a defined function.

```python
class IterIndexState(rx.State):
    color: list[str] = [
        "red",
        "green",
        "blue",
    ]


def create_button(color: rx.Var[str], index: int) -> rx.Component:
    return rx.button(
        f"{index + 1}. {color}",
        padding_y="0.5em",
    )


def enumerate_foreach():
    return rx.vstack(
        rx.foreach(
            IterIndexState.color,
            create_button,
        ),
    )
    ```

docs/components/props.md
tgberkeley: don't see any change here?

docs/components/rendering_iterables.md
tgberkeley: I think for this enumerating iterables example we should show both the examples with the defined function and the lambda so people know that they are identical and that they can do either

docs/components/rendering_iterables.md
tgberkeley: we are addressing this elsewhere so I think we can ignore this comment for now

docs/components/props.md
ElijahAhianyo: When you say define, what do you mean? I put out an explanation on line 39 explaining what class names will do:

> In the example above, the `class_name` prop of the `rx.box` component is assigned a list of class names. This means the `rx.box` component will be styled with the CSS classes `class-name-1` and `class-name-2`.

Is this not what we want?

tgberkeley: in this case I meant we could actually add the class-name-1 with the values that it takes, but it is a small point so let's leave it for now

tgberkeley (approved):

PR #1060, opened 2024-10-29
[ENG-3593]Replace traces of `rx.chakra` in the docs

ElijahAhianyo:
Users should not see `rx.chakra` in the docs


pcweb/flexdown.py
```diff
@@ -214,7 +214,7 @@ def render(self, env) -> rx.Component:
                 *[
                     rx.fragment(
                         rx.text(
-                            rx.chakra.span(
+                            rx.el.span(
```
picklelo, line 217: We have `rx.text.span`

picklelo (approved):

PR #1086, opened 2024-11-13
api for ag grid

tgberkeley:


tgberkeley:
This PR requires a new version of `reflex-ag-grid` to be released!

docs/library/tables-and-data-grids/ag_grid.md
```diff
@@ -561,3 +561,144 @@ The following props are available for `column_defs` as well as many others that
...
+        rx.button("Select All", on_click=my_api.select_all()),
+        rx.button("Deselect All", on_click=my_api.deselect_all()),
+        spacing="4",
+    )
+```
+
+
+The react code for the `select_all()` event handler is `selectAll = (source?: SelectionEventSourceType) => void;`. 
+
+To use this in Reflex as you can see, it should be called in snake case rather than camel case, and as it returns `void` this means that the function should be called with no arguments.
```
masenf, line 613: The `void` means it doesn't return anything. The `source?` indicates that it takes an optional `source` argument

docs/library/tables-and-data-grids/ag_grid.md
```diff
@@ -561,3 +561,144 @@ The following props are available for `column_defs` as well as many others that
...
 - `cell_editor`: `AGEditors | str | None`: Provide your own cell editor component for this column's cells. (Check out the Editing section of this page for more information)
 - `cell_editor_params`: `dict[str, list[Any]] | None`: Params to be passed to the cellEditor component.
+
+
+
+## Functionality you need is not available in Reflex
+
+Since Reflex AG Grid is wrapping the underlying AG Grid library, there is much more functionality available that is currently not exposed in Reflex. Check out this [documentation](https://www.ag-grid.com/react-data-grid/reference/) for more information on what is available in AG Grid.
+
+As Reflex does not expose all the functionality of AG Grid, you can use `ag_grid.api()`, which is a property of the `ag_grid` component, to access the underlying AG Grid API. This allows you to access the full functionality of AG Grid.
```
masenf, line 571: > which is a property of the `ag_grid` component

Technically this is also true, but this page describes the `api` class that's hanging off the `ag_grid` _namespace_

docs/library/tables-and-data-grids/ag_grid.md
```diff
@@ -561,3 +561,144 @@ The following props are available for `column_defs` as well as many others that
...
+            row_data=df.to_dict("records"),
+            column_defs=column_defs,
+        ),
+        rx.button("Export", on_click=my_api.export_data_as_csv()),
+        rx.button("Resize Columns", on_click=my_api.size_columns_to_fit()),
+        spacing="4",
+    )
+```
+
+The react code for both of these is shown below. The key point to see is that both of these functions return `void` and therefore should be called with no arguments.
```
masenf, line 659: again here, the `void` is for the return. the arguments accepted are listed in the parentheses

docs/library/tables-and-data-grids/ag_grid.md
tgberkeley: fixed

docs/library/tables-and-data-grids/ag_grid.md
tgberkeley: fixed

docs/library/tables-and-data-grids/ag_grid.md
tgberkeley: fixed

PR #1089, opened 2024-11-14
[GTM-838][GTM-841]Add template name and fix broken template links

ElijahAhianyo:


ElijahAhianyo:
Screenshot 2024-11-14 at 4 40 35 PM


Initial proposal on what the template cards look like

ElijahAhianyo:
> Screenshot 2024-11-14 at 4 40 35 PM
> Initial proposal on what the template cards look like. Thoughts on this? @Alek99 @picklelo

I kept it this way because the template names are kebab cased on reflex init and I thought it was intuitive to have titles as Title-cases with whitespaces instead of hyphens(or underscores). We could argue that names could simply be formatted for the titles, but that means the name displayed on the title will differ from that when you click the copy button which could be confusing, so instead,  having the template name explicitly stated should clear this up. My only issue here is the use of `Template name` again before specifying the actual name. Not a designer, so would need your feedback on that as well

ElijahAhianyo:
This PR now depends on https://github.com/reflex-dev/templates/pull/44 and https://github.com/reflex-dev/templates/pull/45

pcweb/components/code_card.py
```diff
@@ -2,7 +2,7 @@
 from pcweb.components.icons.icons import get_icon
 
 
-def install_command(command: str) -> rx.Component:
+def install_command(command: str, **props) -> rx.Component:
```
picklelo, line 5: Let's remove the dollar sign in the template

ElijahAhianyo:
Screenshot 2024-11-15 at 7 38 54 PM

ElijahAhianyo:
> Screenshot 2024-11-15 at 7 38 54 PM

@picklelo @tgberkeley lmk what you think

Alek99:
Lets change the init with template

Says in terminal

"""Go to the templates page (http://reflex.dev/templates/) and copy the command to init with a template

The command will be like the following 'reflex init --template template_name"
"""


Then cuts the command and does not take then to the template page, that way they can see the command

Now when they go to template 

When clicking copy on the template page it should copy the whole command

reflex init --template ....

The user pastes this command into their terminal


---------------


The ui will look like this

Screenshot 2024-11-20 at 9 41 18 PM

ElijahAhianyo:
Screenshot 2024-11-24 at 11 05 36 AM

picklelo (approved):

Alek99 (approved):

PR #1104, opened 2024-11-26
nginx.conf tweaks: return 404 page when not found

masenf:
avoid returning a 500 error due to internal rewrite loop, which blocks/breaks dynamic routes

### DIFF
--- a/blog/2024-10-8-self-hosting-reflex-with-docker.md
+++ b/blog/2024-10-8-self-hosting-reflex-with-docker.md
@@ -205,6 +205,7 @@ server {
  server_name frontend;
 
 
+ error_page   404  /404.html;
  error_page   500 502 503 504  /50x.html;
  location = /50x.html {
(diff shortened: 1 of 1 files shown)


blog/2024-10-8-self-hosting-reflex-with-docker.md
```diff
@@ -205,6 +205,7 @@ server {
  server_name frontend;
 
 
+ error_page   404  /404.html;
```
Kastier1, line 208:
```suggestion
 error_page   404  /usr/share/nginx/html/404.html;
```

masenf: we want the app's `/404.html` page because that's where the dynamic route client side redirect logic lives

Kastier1 (approved):

Kastier1 (approved):

PR #1115, opened 2024-12-03
Use `on_mount` for plotly and pyplot state examples

ElijahAhianyo:


docs/library/graphing/other-charts/pyplot.md
```diff
@@ -97,6 +99,7 @@ class PyplotState(rx.State):
     num_points: int = 100
     plot_data: tuple = tuple(np.random.rand(2, 100) for _ in range(3))
```
Alek99, line 100: We should have this as empty in the initial state

Alek99 (changes requested):

PR #1125, opened 2024-12-04
followup aggrid state opti

Lendemor:


picklelo (commented):
Should we split it up so that the code they see still has it in the default, but underlying we load it on load? I don't want to make the examples more complex just for our performance

Alek99:
I'm going to approve for now we can follow up with this

Alek99 (approved):

PR #1127, opened 2024-12-04
Add new pixel

picklelo:


Alek99 (changes requested):
Can we remove [pixels_common_room.py](https://github.com/reflex-dev/reflex-web/blob/653a6f5b0d6e89532c06c256c74f4a290572600e/pcweb/telemetry/pixels_common_room.py)

or any other pixels we dont use with this pr too

Alek99 (approved):

PR #1159, opened 2024-12-19
update upload docs and links to audio, video and image docs

tgberkeley:


masenf (approved):
having the animation as a webm would be optimal, but not blocking on it

PR #1179, opened 2025-01-18
update year in footer automatically

truongd3:


Alek99:
I dont think we want to to be that dynamic, a better solution could be to just refactor it to a constant

PR #1183, opened 2025-01-21
remove posthog pixel

tgberkeley:


Alek99 (changes requested):
remove clear bit swell

Alek99 (approved):

PR #1198, opened 2025-01-29
replace pricing calculator with compute table

carlosabadia:


docs/hosting/regions.md
```diff
@@ -0,0 +1,103 @@
...
+Below is an example of how to deploy your app in several regions:
+
+```bash
+reflex deploy --project f88b1574-f101-####-####-5f########## --region sjc --region iad
+```
+
+By default all apps are deloyed in `sjc` if no other regions are given. If you wish to deploy in another region or several regions you can pass the `--region` flag (`-r` also works) with the region code. Check out all the regions that we can deploy to below:
+
+
+## Config File
```
tgberkeley, line 80: This section about the config file should probably be it's own page

docs/hosting/deploy-quick-start.md
```diff
@@ -72,7 +72,7 @@ If you go back to the Cloud UI you should be able to see your deployed app and o
 
 ```md alert info
 # Setup a Cloud Config File
-To create a `config.yml` file for your app to set your app configuration check out the [Cloud Config Docs]({docs.hosting.app.path}#config-file).
+To create a `config.yml` file for your app to set your app configuration check out the [Cloud Config Docs]({docs.hosting.app_management.path}).
```
tgberkeley, line 75: this path is incorrect now, this is not where the config file is, it should be its own page perhaps

tgberkeley (approved):

PR #1216, opened 2025-02-19
Pass the actual document filename to flexdown render

masenf:


adhami3310 (approved):

adhami3310:
apparently filename doesn't always exist

masenf:
> apparently filename doesn't always exist

i don't know exactly why, but it passes with https://github.com/reflex-dev/flexdown/pull/15... even though that code doesn't seem to do anything with the document filename. will see if there's something i can do to retain compat with older flexgen versions

masenf:
@adhami3310 so it turned out that flexdown/main had the version listed as 0.1.5a7, even though the code there was _newer_ than the published version 0.1.5a11 (Sep 2024).

I just published 0.1.5a12 which has the requisite changes, and updated `requirements.txt` here.

PR #1229, opened 2025-02-28
Add type hints to Image component documentation

akincan-kilic:
If the source of the image is from the PIL package, the type hint must be "Image.Image". It is easy to mistake it for simply "Image" which results in the following error: TypeError: Unsupported type <module 'PIL.Image'>. This simple type hint on the documentation resolves any confusion.


picklelo (approved):
thanks!

PR #1230, opened 2025-02-28
remove reflex chakra

adhami3310:


masenf (approved):
looks good to me. i pushed a fix for the VideoBlock to style it like the AlertBlock. If that looks okay to you, then i'm good to merge, but @Alek99 or @carlosabadia might want to sign off as well.

(lol at `errors.md` 😂, that was a funny way to spell `.py`)

PR #1246, opened 2025-03-26
Use `uv` to simplify the local setup instructions in the README

TimChild:
Update the **Setup Locally** instructions to use `uv` instead of `venv`.

I'm pretty sure you are all using `uv` now for package management, and it seems like a good idea to recommend to others since it's so convenient to use. 

Note: the one liner `uv run --with-requirements ...` doesn't do hot-reloading (not sure why)... although the hot-reloading wasn't working well for me even with a full venv. I included instructions for methods of setup anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,23 @@ We recommend using a virtual environment with Python 3.10 or higher to run this 
 
     *Open the browser and go to `http://localhost:3000/` to see the website.*
 
+**`uv` alternative**
+
+Run the project with `uv` using:
+
+```bash
+uv run --with-requirements requirements.txt reflex run
+```
+
+or
+
+```bash
+uv venv
+uv pip install -r requirements.txt
+uv run reflex run
+```
+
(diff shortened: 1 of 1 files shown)


README.md
Alek99, line 1: Im not sure about this one do you think uv is popular enough yet for this -  I want it to have what most people know already

TimChild: Definitely that should be the aim. You might be right that it is too soon to remove python venv instructions.

I think uv has gained a lot of popularity recently given it's better support for pyproject.toml than many others, but not sure how widespread it is. 

Maybe the uv one liner could just be included as well since that may be new to some uv users, and I think it's a nice way to quickly run things. But also, maybe not worth thinking about this for now.

I'll adjust the PR in the morning, but also feel free to close if you'd like.

adhami3310:
we are now using pyproject and made the instruction to do `pip install .`

PR #1247, opened 2025-03-26
Update Clerk Auth repo (old recommendation has been unmaintained for a long time)

TimChild:
Updated the recommendation for `Clerk Auth` to the new package I released recently as the old one is no longer maintained and doesn't work well with reflex 0.7+ (or even 0.6+).


Alek99:
thanks!

Issue #1162, opened 2025-01-08
Issue with reflex.dev documentation: hover styling example


tartansandal:
Path: /docs/getting-started/dashboard-tutorial

The code to set hover styling for each user row seems to be broken  

```python
def show_user(user: User):
    """Show a user in a table row."""
    return rx.table.row(
        rx.table.cell(user.name),
        rx.table.cell(user.email),
        rx.table.cell(user.gender),
        style=\{"_hover": 
            {"bg": rx.color("gray", 3)}
        },
        align="center",
    )
```

should be

```python
def show_user(user: User):
    """Show a user in a table row."""
    return rx.table.row(
        rx.table.cell(user.name),
        rx.table.cell(user.email),
        rx.table.cell(user.gender),
        _hover: {"bg": rx.color("gray", 3),
        align="center",
    )
```

PR #1252, opened 2025-03-31
fixes issue #1162

LineIndent:


adhami3310:
the slash is only needed if the braces end in the same line

Alek99 (approved):

PR #1261, opened 2025-04-02
bump reflex>=0.7.5

LineIndent:


Alek99 (approved):

Alek99 (changes requested):
need to make it work on main reflex

PR #1262, opened 2025-04-03
[ENG-5373] site navigation

LineIndent:


pcweb/pages/docs/cloud_cliref.py
```diff
@@ -145,6 +145,6 @@ def generate_docs(source: str):
 for module_name, module_value in modules.items():
     docs = generate_docs(module_value)
     title = module_name.replace("_", " ").title()
-    page_data = docpage(f"/docs/hosting/{module_name}/", title)(docs)
+    page_data = docpage(f"/cloud/{module_name}/", title)(docs)
```
Alek99, line 148: This still should be under docs all the doc pages should be under /docs/

Alek99: I think having under hosting is fine we dont have to change it

Alek99 (approved):

PR #1286, opened 2025-04-14
[ENG-5498] Tailwind docs

carlosabadia:


Alek99 (changes requested):
Can you go into more detail on the dynamic styling section how you cannot contruct the string dynamically from state

Alek99 (approved):

PR #1359, opened 2025-05-12
[ENG-5483] right sidebar on-scroll highlight effect (handles clicks too)

LineIndent:


Kastier1:
I can't run this branch i get `Error: Expected ref to be a function, an object returned by React.createRef(), or undefined/null.`

Kastier1 (approved):

PR #1364, opened 2025-05-15
clean up dependencies

adhami3310:


Alek99 (changes requested):
I think we need to add requests

adhami3310:
> I think we need to add requests

what requests?

Alek99 (approved):

PR #1391, opened 2025-05-28
remove ide vid from assets + YT

LineIndent:


carlosabadia (approved):
Maybe this for a follow up pr

But can we include in the IDE features section the drag and drop file upload to any folder? 

Would be nice to also show all these features in the IDE video (rename, delete, add a new file, etc), not only the edit code and save

Alek99:
> Maybe this for a follow up pr
> 
> But can we include in the IDE features section the drag and drop file upload to any folder?
> 
> Would be nice to also show all these features in the IDE video (rename, delete, add a new file, etc), not only the edit code and save

Lets add a follow up ticket for this

PR #1392, opened 2025-05-29
[ENG-6021] re-arrange the support header to the top section

LineIndent:


carlosabadia (changes requested):
image
Can we add a border the last row?

PR #1393, opened 2025-05-29
[ENG-6018] 4-col price cards

LineIndent:
@carlosabadia there's a bit of a mashed look to the cards between the lg and md breakpoints (which is expected) - i tried making a new break point and extending the tailwind but it was not registering - so I ended up making the cards `min-w-0` so that they can shrink as the screen is re-sized. Do you see any other alternative?

### DIFF
--- a/pcweb/pages/pricing/plan_cards.py
+++ b/pcweb/pages/pricing/plan_cards.py
@@ -153,12 +153,12 @@ def card(
                     price,
                     color_scheme="gray",
                     size="3",
-                    class_name="font-medium text-lg",
+                    class_name="font-medium 2xl:text-lg text-base w-fit",
                 )
                 if price
                 else rx.fragment()
             ),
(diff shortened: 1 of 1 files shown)


carlosabadia:
@LineIndent 
image

I think for smaller breakpoints, we can change tier/price div to

```python
rx.el.div(
    rx.el.h3(title, class_name="font-semibold text-slate-12 text-2xl"),
    (
        rx.badge(
            price,
            color_scheme="violet",
            size="3",
            class_name="font-medium 2xl:text-lg text-base w-fit",
        )
        if price
        else rx.fragment()
    ),
    class_name="flex 2xl:items-center mb-2 2xl:gap-4 gap-2 2xl:flex-row flex-col",
),

<img width="1486" alt="image" src="https://github.com/user-attachments/assets/fe61d7e8-d823-40c1-bb3c-7cdd49bcba42" />

carlosabadia (approved):

PR #1396, opened 2025-05-29
[HOS-682] [HOS-737] more rounded IDE video

LineIndent:


carlosabadia (changes requested):
Video looks good. Could we mention the drag and drop files in the features section?

PR #1413, opened 2025-06-05
[ENG-6115] use case page

LineIndent:


pcweb/pages/use_case/use_case.py
```diff
@@ -0,0 +1,34 @@
...
+from pcweb.pages.framework.index_colors import index_colors
+
+
+document = flexdown.parse_file("pcweb/pages/use_case/use-case.md")
+
+def use_case_content():
+    return rx.box(xd.render(document, document.filename))
+
+
+@rx.page(route="/use-case", title="Use Case - Reflex")
```
tgberkeley, line 17: This should say `@rx.page(route="/use-cases", title="Use Cases - Reflex")`

Alek99:
Can you put it in the doc page  instead
Screenshot 2025-06-05 at 5 35 59 PM

Also we should duplicate the page to be at the end of the getting started in framework section.

Alek99 (changes requested):
Ok we can remove the non doc page for the use cases - I dont like how the logic is duplicated in two exact same files but maybe there is not a good solution right now. 


But we should get rid of the non doc page one and link the navbar to the ai builder page use case

PR #1488, opened 2025-07-10
fix components links

Lendemor:


Alek99:
completed with another pr

PR #1585, opened 2025-08-15
ENG-7371: bring back pricing & change demo url (includes ENG-7392)

LineIndent:


Darksinian (dismissed):
Discussed with the team and they want the following chages:
1. Remove pro tier card completely
2. have a crossed out item in the free titled "Database Integrations" so that database is not possible for free.
3. Either remove the "AI integrations" completely or at least make it "OpenAI Integration" and make it enabled for free since they will be able to use that

carlosabadia (approved):

PR #1605, opened 2025-09-02
ENG-7547: update rxconfig docs

LineIndent:


masenf:
i thought this was supposed to be auto generated, and i think that stopped happening when BaseConfig was split out of the Config object. Will look into the autogen stuff a bit and see if we can bring that back.

PR #1606, opened 2025-09-02
ENG-7551: add booked page

LineIndent:


pcweb/pages/booked.py
```diff
@@ -0,0 +1,46 @@
...
+            "We’ve sent you a confirmation email with all the details.",
+            class_name="font-md text-balance text-slate-9",
+        ),
+        class_name="section-header",
+    )
+
+
+
+
+@rx.page(
```
carlosabadia, line 26: Is there a template we can use? if not lets add the `create_meta_tags` as well

carlosabadia (approved):

PR #1612, opened 2025-09-04
ENG-7586: react flow docs

LineIndent:


adhami3310 (commented):
i updated `reflex-enterprise` to have react flow so we can do `demo` now

we can either ignore the custom node/edges until we get another reflex version or you can also upgrade reflex to latest main

PR #1649, opened 2025-10-03
ENG-7879: integration page for web

LineIndent:


Alek99 (changes requested):
As discussed on friday the integration page need to be within the sidebar etc.

Also a good chunk of these integrations should use the https://simpleicons.org/?q=goggle

I suggest we put this in the shared reflex-ui repo so the implementation can be used across projects

